PR #101, opened 2026-07-07
mantis: add SQLi and SSRF sinks to source-sink heuristic

deonmenezes:
## What gap this closes

`.codex/agents/recon.toml` explicitly instructs the Recon agent to treat "SQL/NoSQL query construction" and "HTTP client calls (SSRF)" as sinks of interest. But the actual heuristic source/sink tagger that feeds Recon/Detect (`.codex/mcp-servers/program-analysis/source_sink_rules.js`) had:

- **Zero SQL-injection sink patterns** for JS, Python, or Go — only Java (`statement.execute`) was covered.
- **Zero SSRF sink patterns** for any language at all.

That's a real mismatch between what the pipeline's own docs say it looks for and what the tool could actually flag — a coverage gap for two of the vulnerability classes this run's scan is specifically asked to check for (SQL injection is named directly).

## What changed

Added sink rules (all CWE-tagged, `kind: "sink"`, additive only — no existing rules touched):

- **SQLi**: `js.sql.query_template_literal`, `js.sql.query_string_concat`, `py.sql.execute_fstring`, `py.sql.execute_format_concat`, `go.sql.query_sprintf` (CWE-89)
- **SSRF**: `js.ssrf.dynamic_url`, `py.ssrf.dynamic_url`, `go.ssrf.dynamic_url` (CWE-918)

Patterns are scoped to the *dynamic construction* case (template-literal interpolation, string concatenation, `%`-format, `f`-strings, `fmt.Sprintf`) so parameterized queries (`db.query("... WHERE id = ?", [id])`) and static URLs (`fetch("https://api.example.com/health")`) are **not** flagged — keeping the recall gain from inflating false positives, consistent with this heuristic's "detect generous, but don't manufacture noise" role feeding the ruthless Validate stage downstream.

## Testing

No existing test harness covers this pure-data rules file. Verified manually:
- `node --check` passes (syntax).
- Ran the new + existing rules against 15 hand-written snippets (vulnerable and safe variants of each new pattern, across JS/Python/Go) — all matched or correctly did not match as expected.
- Ran the full `source_sink_rules.js` against a small sample Flask handler containing an f-string SQL query and an f-string SSRF call — both new sinks fired with correct line/CWE attribution, alongside the existing `py.flask.request` source detections.
- `prettier --check`/`--write` applied to match repo formatting conventions.

---
_Generated by [Claude Code](https://claude.ai/code/session_0114JYpExynyBDxWaNGxYTRH)_

### DIFF
--- a/.codex/mcp-servers/program-analysis/source_sink_rules.js
+++ b/.codex/mcp-servers/program-analysis/source_sink_rules.js
@@ -79,6 +79,29 @@ const RULES = [
     pattern: /\b(node-serialize|serialize\.unserialize)\s*\(/g,
     cwe: "CWE-502",
   },
+  {
+    lang: "js",
+    kind: "sink",
+    id: "js.sql.query_template_literal",
+    pattern: /\.(query|execute)\s*\(\s*`[^`]*\$\{/g,
+    cwe: "CWE-89",
+  },
+  {
+    lang: "js",
+    kind: "sink",
+    id: "js.sql.query_string_concat",
+    pattern:
+      /\.(query|execute)\s*\(\s*(['"][^'"]*['"]\s*\+|[A-Za-z_$][\w$]*\s*\+\s*['"])/g,
+    cwe: "CWE-89",
+  },
+  {
+    lang: "js",
+    kind: "sink",
+    id: "js.ssrf.dynamic_url",
+    pattern:
+      /\b(axios(\.\w+)?|fetch|got)\s*\(\s*(`[^`]*\$\{|[A-Za-z_$][\w$.]*\s*\+\s*)/g,
+    cwe: "CWE-918",
+  },
 
   // Python
   {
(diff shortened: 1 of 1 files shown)


deonmenezes:
**Re: the failing "CI required" check** (cargo-deny + sdk).

Root cause: `cargo-deny`'s `advisories` check fails on `RUSTSEC-2026-0204` (invalid pointer dereference in `crossbeam-epoch`'s `fmt::Pointer` impl), pinned at `crossbeam-epoch v0.9.18` in `codex-rs/Cargo.lock` via `crossbeam-deque` → `ignore` → `codex-file-search`. This is pre-existing and unrelated to this PR's diff (a JS detection-rules data file only) — it would fail identically on `main` today. (`sdk` shows `cancelled`, which is a downstream effect of the same required-check job failing, not a separate issue.)

Unlike this repo's other `deny.toml` advisory exceptions (all documented as "no fixed release yet"), this one has an available fix: `crossbeam-epoch >= 0.9.20`. So the right fix is a version bump, not adding it to the ignore list.

I attempted `cargo update -p crossbeam-epoch --precise 0.9.20` from this sandboxed session but couldn't complete it: this workspace's `cargo update` always refreshes all pinned sources first, including the unrelated git submodule `codex-realtime-webrtc` → `libwebrtc` (`rust-sdks.git`) → `libyuv` (`chromium.googlesource.com`), and that host is blocked by this environment's network egress policy. I didn't hand-edit `Cargo.lock`'s checksum blind since I can't verify it without network access.

Suggested fix for someone with full network access:
```
cd codex-rs
cargo update -p crossbeam-epoch --precise 0.9.20
cargo deny check advisories
```
Then commit the resulting `Cargo.lock` diff (should be a single-crate version bump). Happy to do this myself if this session's network access is widened, otherwise this needs a maintainer with a normal dev environment.

---

PR #120, opened 2026-07-11
detect: normalize codeql_analyze severity to the shared finding vocabulary

deonmenezes:
## What gap this closes

Every other Detect-stage server normalizes its tool-native severity into the shared `info/low/medium/high/critical` vocabulary that `mantis_findings` enforces (`VALID_SEVERITIES` in `.codex/mcp-servers/findings/server.js`): semgrep maps `ERROR/WARNING/INFO`, bandit maps `HIGH/MEDIUM/LOW`, trivy maps `CRITICAL/HIGH/MEDIUM/LOW/UNKNOWN`. `codeql_analyze` (`.codex/mcp-servers/codeql/server.js`) was the one holdout: it only ever returned the raw SARIF `level` (`error`/`warning`/`note`), which is not a legal `finding_create`/`finding_update` severity and also ignores CodeQL's own per-rule `security-severity` CVSS-like score (a much more accurate signal than the 3-value `level`). In practice this meant an agent taking a CodeQL finding straight to `finding_create` either had to guess a severity (inconsistent with the other five detectors) or pass the raw `level` and get it rejected.

No overlap with any currently open detection PR from this routine — this is scoped entirely to `.codex/mcp-servers/codeql/`, which none of the other open PRs touch.

## What changed

- Extracted the new mapping logic into a pure, side-effect-free module, `.codex/mcp-servers/codeql/severity.js`, so it's unit-testable without importing `server.js` (which starts listening on stdin as soon as it's required — the reason none of the other server.js files have direct unit tests today).
- `severityForResult(result, ruleSeverityIndex)`: prefers the rule's `security-severity` score (bucketed at the same 9/7/4 thresholds GitHub code scanning itself uses: critical/high/medium/low), falls back to the rule's `problem.severity`, then falls back to mapping the SARIF `level` (`error`→high, `warning`→medium, `note`→low, default→info — matching the existing `level` default of `"warning"`).
- `buildRuleSeverityIndex(run)`: indexes a SARIF run's `tool.driver.rules` and `tool.extensions[].rules` by rule id so each result can look up its own rule's severity metadata.
- `codeqlAnalyze()` now attaches a `severity` field to every finding alongside the existing `level`, ready to pass straight into `finding_create`/`finding_update`.
- Updated the tool description to document the new field.

Additive only: the existing `level`, `rule_id`, `message`, `path`, `start_line` fields are unchanged.

## Testing

```
node --check .codex/mcp-servers/codeql/server.js
[...]

PR #121, opened 2026-07-11
detect: add SSRF (CWE-918) sink coverage to source_sink_scan

deonmenezes:
## What gap this closes

`source_sink_scan`'s rule table (`.codex/mcp-servers/program-analysis/source_sink_rules.js`) had sink coverage for command injection (CWE-78), XSS (CWE-79), and deserialization (CWE-502) across JS/Python/Go/Java, but **no rules for Server-Side Request Forgery (CWE-918)**. A request-controlled value flowing straight into an outbound HTTP call (`fetch(req.query.url)`, `axios.get(req.body.callbackUrl)`, `requests.get(request.args["url"])`, `http.Get(r.FormValue("target"))`, `new URL(request.getParameter("target"))`) never surfaced as a `candidate` for Detect to register, since no rule tagged that shape as a sink. SSRF is a common, high-impact class (cloud metadata endpoint access, internal network pivoting) that scanners like semgrep/CodeQL may also catch, but this heuristic recall layer had a blind spot for it specifically.

## What changed

Added 7 new sink rules to `RULES`, all tagged `CWE-918`, `kind: "sink"`:

- JS: `js.ssrf.fetch_source`, `js.ssrf.axios_source`, `js.ssrf.http_request_source`
- Python: `py.ssrf.requests_source`, `py.ssrf.urlopen_source`
- Go: `go.ssrf.http_get_source`
- Java: `java.ssrf.url_connection_param`

Each rule requires a request-source term (`req.query`/`req.body`/..., `request.args`/..., `r.FormValue(...)`, `request.getParameter(...)`) to appear *inside* the call parens, so a static/hardcoded URL passed to the same function does not fire -- mirrors the pattern already used for the CWE-22 path-traversal sinks on other open branches.

Also updated `.codex/skills/program-analysis/SKILL.md`'s description of `source_sink_scan` to mention the new SSRF sink coverage, and added `source_sink_rules.test.js` (14 `node:test` cases: one fire + one no-fire case per new rule, plus a CWE-918 count/uniqueness check and a global rule-id uniqueness check) since none existed for this file yet.

## Why it's safe

- Pure additive change to a heuristic regex rule table -- no changes to the scanner's control flow, no new dependencies.
- `source_sink_scan` only ever emits `candidate`s (never a severity or a confirmed finding); per the findings-spine skill, each hit still has to be traced for reachability and attacker-simulated before it can become `confirmed`. This only widens Detect's recall, consistent with "Detect is generous, Validate is ruthless."
- Verified with `node --test .codex/mcp-servers/program-analysis/source_sink_rules.test.js` (14/14 passing) and `prettier --check` on both changed/added JS files.

## Test plan

- [x] `node --test .codex/mcp-servers/program-analysis/source_sink_rules.test.js` -- 14/14 pass
- [x] `node -c` syntax check on both JS files
- [x] `npx prettier --check` on both changed/added JS files

---
_Generated by [Claude Code](https://claude.ai/code/session_01JYBnjVcPUJ8emUjngjC5wM)_


deonmenezes:
Investigated both failing checks — neither is caused by this diff (which only touches `.codex/mcp-servers/program-analysis/source_sink_rules.js`, its new test file, and `.codex/skills/program-analysis/SKILL.md`):

- **Vercel**: fails identically on every other open PR regardless of content (confirmed on #108, #120) — a pre-existing, repo-wide deployment integration issue for an unrelated "opentradex-dashboard" project.
- **cargo-deny / cargo-deny**: fails on `RUSTSEC-2026-0204` (invalid pointer dereference in `crossbeam-epoch` 0.9.18's `fmt::Pointer` impl, fixed in >=0.9.20), a deep transitive dependency pulled in via `crossbeam-deque`/`rayon` in `codex-rs/Cargo.lock`. Also fails identically on #120, so it's a pre-existing advisory against `main`'s lockfile, not something this PR's diff introduced. Bumping `crossbeam-epoch` is an unrelated Rust-dependency fix that belongs in its own PR rather than bundled here.

No action taken on either in this PR; flagging so review isn't blocked on unrelated red checks.

---

deonmenezes:
Note for reviewers: an unrelated automated run briefly pushed a different SSRF-coverage attempt onto this same branch name (`mantis-detection/ssrf-sink-coverage`), which force-overwrote this PR's commit for a few minutes. That was a mistake — this branch's original commit (`dc05f49`) has been restored verbatim, so this PR is back to its original content (source-gated SSRF rules + unit tests). No action needed here; the other attempt will go out under a different branch/PR instead.

---

PR #122, opened 2026-07-11
detect: add C#/.NET source/sink coverage to source_sink_scan

deonmenezes:
## What gap this closes

`source_sink_scan` (the heuristic recall pass in `.codex/mcp-servers/program-analysis/`) only had rules for JS/TS, Python, Go, and Java. Any C#/.NET codebase scanned by Mantis produced zero source/sink candidates — not because C# targets are safe, but because the tool had no rules for the language at all. That's a coverage gap, not a clean bill of health.

## What changed

- `source_sink_rules.js`: added a `cs` rule set —
  - sources: ASP.NET `Request.(Query|Form|Headers|Cookies|QueryString)`, `Environment.GetCommandLineArgs()`
  - sinks: `Process.Start(` (CWE-78, command injection), `new SqlCommand(` (CWE-89, SQL injection), `Response.Write(` and `Html.Raw(` (CWE-79, XSS), `new BinaryFormatter(` (CWE-502, insecure deserialization)
- `server.js`: registered the `.cs` extension in `EXTENSION_TO_LANG` and added `"cs"` to the `source_sink_scan` `languages` enum.
- `program-analysis` skill doc: updated the language list mentioned for `source_sink_scan`.

Followed the same shallow instantiation-site heuristic already used for the Java rules (e.g. `java.objectinputstream` flags `new ObjectInputStream(` itself, not the later `.readObject()` call) — consistent with this tool's stated job: a fast recall proxy for the Detect/Validate pipeline to trace and connect, not a taint engine.

## Why

Silent gaps in language coverage are worse than false positives here: a scan that reports "0 findings" on a C# target reads as "nothing to see," when really the tool just can't see that language. This closes that specific blind spot without touching any other language's rules or the scan's semantics.

## Testing

- `node --check` on both modified JS files.
- `prettier --check` on all three changed files (all pass).
- Ran the new `cs` rules directly against a synthetic C# snippet exercising every new source/sink; all six fired at the expected line with the expected CWE.

No existing tests cover this heuristic module (there is no test harness under `.codex/mcp-servers/`), consistent with the other language rule sets already in this file.

---
_Generated by [Claude Code](https://claude.ai/code/session_01ASEcJZCbPdwupzqxhNxfNj)_


deonmenezes:
The `cargo-deny / advisories` check failure here is unrelated to this PR's diff (which only touches JS files under `.codex/mcp-servers/program-analysis/` and one skill doc). It's a newly-published advisory, [RUSTSEC-2026-0204](https://rustsec.org/advisories/RUSTSEC-2026-0204), against `crossbeam-epoch v0.9.18`, a transitive dependency pulled in via `ignore` → `codex-file-search` deep in the Rust `codex-cli` tree. It would fail the same way on `main` and on any other open PR right now — fixing it means bumping that transitive dependency repo-wide, which is out of scope for a detection-rules change and best done as its own PR. Not fixing it here to avoid scope creep on an unrelated dependency.

---

deonmenezes:
Re-ran the failed jobs once. `cargo-deny` failed again for the same pre-existing reason noted above (unrelated). `sdk / sdks` was cancelled again, both times at almost exactly the ~30 minute mark during the "Build codex with Cargo" step (16:08:11→16:37:56 on attempt 1, 16:39:11→17:09:00 on attempt 2) — that's a systemic CI job timeout on the full Rust workspace build, not flakiness, and unrelated to this PR's two-file JS diff. Not re-triggering again since a third run would hit the same timeout; this looks like it needs attention on the `sdk` workflow's timeout/build-time budget independent of this PR.

---

PR #127, opened 2026-07-12
Add SQL-injection sink rules to the source/sink heuristic scanner

deonmenezes:
## What gap this closes

`program-analysis/source_sink_rules.js` is the dependency-free heuristic source/sink tagger the `detector` and `recon` agents fall back on when semgrep/CodeQL aren't available (or to widen recall alongside them). It had **zero CWE-89 (SQL injection) sink signatures for JavaScript, Python, or Go** — only Java's generic `statement.execute()` call was covered. That's a real coverage gap: SQLi is one of the vulnerability classes the Mantis pipeline is explicitly meant to catch, and it's completely invisible to this scanner for three of the four supported languages.

## What changed

Added 7 new sink rules (all `CWE-89`), one per unparameterized-query-construction pattern per language:

- **JS/TS**: `.query()`/`.execute()` called with a template literal containing `${...}` interpolation, or with string concatenation (`+`)
- **Python**: `.execute()` called with an f-string, `%`-formatting, string concatenation, or a `.format()` call
- **Go**: `.Query()`/`.QueryRow()`/`.Exec()` called with string concatenation, or with `fmt.Sprintf(...)`

Each rule targets the actual injection *pattern* (unparameterized string construction at the call site) rather than flagging every DB call, so it stays high-signal instead of drowning the Detect stage in trivial candidates from properly-parameterized queries.

## Testing

Added `test_source_sink_rules.js`, a standalone smoke test (no framework, run via `node`) that asserts each new rule:
1. fires on a representative vulnerable snippet, and
2. does **not** fire on the equivalent parameterized/safe call

```
$ node .codex/mcp-servers/program-analysis/test_source_sink_rules.js
ok - 7 SQL-injection sink rules verified
```

Also ran `prettier --check` on both touched files (repo's existing `format` script covers `**/*.js`).

## Scope

Small, additive, read-only change to detection signatures only — no changes to the MCP server's request/response shape, no new dependencies, no changes to any other pipeline stage.

---
_Generated by [Claude Code](https://claude.ai/code/session_01D8yeWdzwiGs3JRcCATVkof)_

### DIFF
--- a/.codex/mcp-servers/program-analysis/source_sink_rules.js
+++ b/.codex/mcp-servers/program-analysis/source_sink_rules.js
@@ -79,6 +79,20 @@ const RULES = [
     pattern: /\b(node-serialize|serialize\.unserialize)\s*\(/g,
     cwe: "CWE-502",
   },
+  {
+    lang: "js",
+    kind: "sink",
+    id: "js.sql.query_template_literal",
+    pattern: /\.(query|execute)\s*\(\s*`[^`]*\$\{/g,
+    cwe: "CWE-89",
+  },
+  {
+    lang: "js",
+    kind: "sink",
+    id: "js.sql.query_string_concat",
+    pattern: /\.(query|execute)\s*\(\s*['"][^'"]*['"]\s*\+/g,
+    cwe: "CWE-89",
+  },
 
   // Python
   {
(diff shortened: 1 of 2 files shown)


deonmenezes:
The `cargo-deny` failure here is pre-existing and unrelated to this diff (which only touches two files under `.codex/mcp-servers/program-analysis/`, no `Cargo.toml`/`Cargo.lock` changes). It's the `advisories` check failing on a newly-published advisory:

- **RUSTSEC-2026-0204** in `crossbeam-epoch v0.9.18` (pulled in transitively via `ignore` → `crossbeam-deque` → `codex-file-search`). Fix upstream is `cargo update -p crossbeam-epoch` (>=0.9.20).

This would fail the same way on `main` or any other PR right now since it's a lockfile-wide advisory, not something introduced here. I'm leaving it out of this PR to keep the diff focused on the detection-rule change — recommend bumping `crossbeam-epoch` in a separate, dedicated dependency-update PR.

---

PR #130, opened 2026-07-13
fix: redact generically-named secrets in JSON-shaped http_audit bodies

deonmenezes:
## What gap this closes

`mantis_http_audit` (`.codex/mcp-servers/http-audit/server.js`) is the DAST evidence tool used to turn a captured HTTP request/response into a bounded, redacted evidence pack before it's attached to a finding. Its stated invariant (see `SKILL.md` for `http-evidence`) is that raw secrets must never land in a finding, artifact, report, or prompt.

Its generic secret-key redaction pattern only matched form/query-encoded pairs:

```
token=abc123&other=1   -> token=[REDACTED]&other=1   (caught)
```

But it never matched JSON-object shaped bodies, which is the dominant format for modern API request/response bodies:

```
{"username":"bob","password":"hunter2","api_key":"sk_live_..."}
```

Before this change, that whole body passed through `http_audit` untouched — `password` and `api_key` values leaked verbatim into the evidence pack (verified locally before the fix; see reproduction below). Only the pre-existing shape-based patterns (AWS keys, GitHub/Slack tokens, JWTs, PEM private keys) would still catch specific credential formats; a generically-named JSON secret field with no distinctive shape (a plain password, an opaque session token, etc.) had no coverage at all.

## The fix

Added a second generic pattern alongside the existing form/query one, matching `"key": "value"` / `key: 'value'` JSON/JS-object shapes for the same generic secret-key names (`token`, `api_key`, `secret`, `password`, `auth`, `session_id`, `sig`, `signature`, ...), redacting only the value while preserving the surrounding quoting style.

## Verification

Reproduced the gap and confirmed the fix with a standalone Node smoke test exercising the full `SECRET_VALUE_PATTERNS` list:

- `{"password":"hunter2","api_key":"sk_live_..."}` → both values now redacted, `username` (non-sensitive key) left untouched
- Nested JSON (`{"auth": {"token": "abc.def.ghi"}}`) → redacted
- Single-quoted JS-object style → redacted
- Pre-existing form/query encoding (`token=abc123&other=1`) → unaffected, still redacted
- Non-secret fields (`"note"`, `"count"`) → left untouched (no over-redaction)

Ran `prettier --check` on the changed file — passes. No existing automated test suite covers these pure-Node MCP servers (none exist yet for this directory), so verification was manual per above.

## Scope

Single, focused change to one redaction pattern list in one file. No API/schema changes, no new dependencies.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FxzAXjtCi18cvumQSFHeCQ)_

### DIFF
--- a/.codex/mcp-servers/http-audit/server.js
+++ b/.codex/mcp-servers/http-audit/server.js
@@ -49,6 +49,15 @@ const SECRET_VALUE_PATTERNS = [
     /\b(token|api[_-]?key|access[_-]?token|refresh[_-]?token|secret|password|passwd|auth|session[_-]?id|sig|signature)=([^&\s]+)/gi,
     (_match, name) => `${name}=[REDACTED]`,
   ],
+  // Same generically-named secret keys, but in JSON/JS-object body shape
+  // ("key": "value" or key: 'value') instead of form/query encoding -- JSON
+  // is the dominant modern API body format and was previously missed
+  // entirely, so a JSON request/response body sailed through un-redacted.
+  [
+    /(["']?)\b(token|api[_-]?key|access[_-]?token|refresh[_-]?token|secret|password|passwd|auth|session[_-]?id|sig|signature)\1\s*:\s*(["'])[^"']*\3/gi,
+    (_match, keyQuote, name, valueQuote) =>
+      `${keyQuote}${name}${keyQuote}: ${valueQuote}[REDACTED]${valueQuote}`,
+  ],
 ];
 
 const MAX_BODY_PREVIEW = 512;


deonmenezes:
`cargo-deny / cargo-deny` is failing on `advisories`, but it's unrelated to this PR's diff (a pure Node.js file under `.codex/mcp-servers/http-audit/`, no `Cargo.toml`/`Cargo.lock` touched).

Root cause: a newly-published advisory, [RUSTSEC-2026-0204](https://rustsec.org/advisories/RUSTSEC-2026-0204) (invalid pointer dereference in `crossbeam-epoch`'s `fmt::Pointer`/`fmt::Display` impls), now flags `crossbeam-epoch 0.9.18`, which is already pinned on `main`'s `codex-rs/Cargo.lock` (pulled in transitively via `ignore` → `crossbeam-deque` → `codex-file-search`, deep in the workspace). This would fail `cargo-deny` on any PR right now regardless of what it changes — it's a pre-existing, repo-wide advisory gate failure, not something introduced here.

Fixing it means bumping `crossbeam-epoch` (`cargo update -p crossbeam-epoch` to `>=0.9.20`) and validating the whole Rust workspace still builds/tests — out of scope for this small, focused secret-redaction fix, so I'm leaving it out of this PR rather than bundling an unrelated, riskier dependency bump. Flagging here for visibility; may be worth a dedicated PR.

---

PR #131, opened 2026-07-13
docs: log this run's blocked scan attempt against vulnerable-bounty-site

deonmenezes:
## Context

Part of the recurring 4h maintenance+testing routine, which is authorized to run a discovery-only scan against `https://vulnerable-bounty-site.vercel.app`.

## What happened this run

No scan was actually possible: every outbound HTTPS request from this session's environment was rejected at the network egress proxy with a policy-level 403 (`gateway answered 403 to CONNECT (policy denial or upstream failure)`), confirmed via the proxy's own status endpoint. This wasn't specific to the target — the same rejection happened for unrelated control domains (`example.com`, `vercel.app`, `nextjs.org`), so it's a categorical "this environment has no general internet egress" policy, not a target-side block. No request ever reached the target, so there are no scan findings this run.

## What this PR adds

`reports/scans/vulnerable-bounty-site.md` — a small running log for this target, starting with today's entry documenting the network block and what needs to change (the environment's outbound network policy needs to allow `vulnerable-bounty-site.vercel.app`, or general internet egress, before this routine can actually reach it). Future runs can append to this file so scan results are diffable over time instead of starting cold each cycle.

## Scope

Docs-only addition (new file under `reports/`), no code changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FxzAXjtCi18cvumQSFHeCQ)_

### DIFF
--- a/reports/scans/vulnerable-bounty-site.md
+++ b/reports/scans/vulnerable-bounty-site.md
@@ -1,0 +1,33 @@
+# Scan log — vulnerable-bounty-site.vercel.app
+
+Authorized discovery-only target for the recurring "maintenance + testing"
+routine (fires every 4h). This file tracks scan attempts/findings over time
+so each run can diff against the last one instead of starting cold.
+
+Target: `https://vulnerable-bounty-site.vercel.app`
+Scope: discovery-only (no active exploitation, no destructive requests).
+
+---
+
+## 2026-07-13T00:08Z — blocked at the network layer, no scan performed
+
+Every outbound HTTPS request from this session's environment — to the
+target *and* to unrelated control domains (`example.com`, `vercel.app`,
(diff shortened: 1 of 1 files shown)


deonmenezes:
Same `cargo-deny / cargo-deny` failure as #130, and it confirms that one's diagnosis: this PR only adds a markdown file under `reports/scans/` — no Rust/Cargo files touched at all — so the failure is definitely repo-wide/pre-existing on `main` ([RUSTSEC-2026-0204](https://rustsec.org/advisories/RUSTSEC-2026-0204) in `crossbeam-epoch`, already pinned in `codex-rs/Cargo.lock`), not something either PR introduced. Not fixing it here either, for the same out-of-scope reasoning as #130.

---

PR #139, opened 2026-07-16
detect: add SQL injection sinks to source_sink_scan (JS/Python/Go)

deonmenezes:
## Summary

Recurring detection-maintenance pass. `mantis_program_analysis`'s `source_sink_scan` heuristic (`.codex/mcp-servers/program-analysis/source_sink_rules.js`) had a real coverage gap: CWE-89 (SQL injection) sinks were only defined for **Java** (`statement.execute`). JS, Python, and Go targets got **zero** SQLi candidates from this tool despite SQLi being an OWASP top-tier vuln class — a significant false-negative gap for a security scanner.

This adds SQLi sink patterns for the three missing languages:

- **JS/TS**: `.query(`/`.execute(` called with a template literal containing `${...}` interpolation, or with string concatenation (`"..." + var`)
- **Python**: `.execute(`/`.executemany(`/`.raw(` called with an f-string, or with `%`-formatting / `.format()` on a literal
- **Go**: `.Query`/`.QueryRow`/`.QueryContext`/`.Exec`/`.ExecContext` called with `fmt.Sprintf(...)`

Rather than flagging every DB-call site (which would flood output with false positives on parameterized queries — the safe and common case), these patterns key off the actual vulnerable shape: dynamic string construction flowing straight into the query call. Verified this precision manually: parameterized calls like `conn.query("... WHERE id = ?", [id])`, `cursor.execute("... %s", (id,))`, and `db.Query("... $1", id)` do **not** match.

Also touched `.codex/skills/program-analysis/SKILL.md` to mention the new sink coverage in its one-line description of `source_sink_scan`.

## Why

`source_sink_scan` is documented as a *recall* tool feeding the Detect/Recon stage of the Mantis pipeline (see `.codex/MANTIS.md` / `.codex/skills/program-analysis/SKILL.md`) — its whole job is to widen the candidate list cheaply before reachability tracing. A recall tool with a blind spot on one of the most common injection classes, across 3 of 4 supported languages, undermines that job. This is a small, additive, well-scoped fix in line with the existing rule style (see the analogous precision tricks already used for `py.subprocess.shell_true` and `py.yaml.load_unsafe`).

## Test plan

- [x] `node --test .codex/mcp-servers/program-analysis/source_sink_rules.test.js` — 11 new unit tests, all passing (positive match on vulnerable snippets, negative match on parameterized/safe equivalents for each new rule)
- [x] Manual end-to-end check: ran the rule set against synthetic `app.js`/`app.py`/`app.go` files with one vulnerable line and one safe parameterized line each — only the vulnerable lines were flagged
- [x] `node --check` on both changed/added JS files (no syntax errors)
- [x] No existing behavior changed — this only adds new rules to the `RULES` array; every existing rule and its regex is untouched

## CI status

`cargo-deny / cargo-deny` is currently failing on this PR, but it's unrelated to this diff — see the PR comment below. It's a newly published RUSTSEC advisory against a pre-existing transitive Rust dependency (`crossbeam-epoch`), and this PR touches no Rust code or dependency files. It will block every PR on this repo (this one and others) until someone bumps that dependency or adds a `deny.toml` exception, which is out of scope here.

## Scan findings (part 2 of this routine's mandate)

Added `reports/2026-07-16_vulnerable-bounty-site-scan.md`: the authorized discovery-only scan against `https://vulnerable-bounty-site.vercel.app` could not run this cycle — both `curl` (via the session's egress proxy) and `WebFetch` got 403s reaching the target, which this environment's proxy diagnostics identify as an organizational egress-policy denial, not a transient error. No scan traffic reached the target. See the report file for details and what's needed before the next attempt.


deonmenezes:
CI note: `cargo-deny / cargo-deny` is currently failing on the `advisories` check with a newly published `RUSTSEC-2026-0204` (invalid pointer dereference in `crossbeam-epoch`'s `fmt::Pointer` impl, fixed in >=0.9.20). This is unrelated to this PR — the diff here only touches `.codex/mcp-servers/program-analysis/*` (JS) and one skill doc, no `Cargo.toml`/`Cargo.lock` changes, and `crossbeam-epoch` is a pre-existing transitive dependency. It's a repo-wide blocking-CI gate, so it'll block every PR (including ones already on `main`) until someone runs `cargo update -p crossbeam-epoch` or adds a `deny.toml` ignore — that's out of scope for this detection-rules change, flagging it separately rather than bundling an unrelated dependency bump here.

---

PR #143, opened 2026-07-17
fix: widen mantis_http_audit secret redaction to close JSON-body + provider-key gaps

deonmenezes:
## Summary

Scheduled maintenance run — one incremental detection/evidence-quality fix in the `mantis_http_audit` MCP server (`.codex/mcp-servers/http-audit/server.js`).

**The gap:** `mantis_http_audit` is the tool every Mantis stage is supposed to route captured HTTP evidence through so that raw secrets never land in a finding, report, or prompt (PRD section 9/11 invariant). Its generically-named-secret pattern (`token`, `password`, `api_key`, `secret`, ...) only matched query/form-encoded `key=value` pairs. The exact same field shaped as a JSON body member — `"password": "hunter2"` — is arguably the more common shape in modern JSON APIs, and it sailed through completely unredacted. That's a real secret-leak path in a tool whose entire job is to guarantee that doesn't happen, and it directly weakens coverage of the "exposed secrets" vulnerability class this run's scan is also checking for.

**The fix:**
- Add a JSON-body variant of the generic secret-field pattern (`"password":"[REDACTED]"` instead of leaking the value).
- Round out shape-based coverage for provider key formats the existing list missed: GitHub fine-grained PATs (`github_pat_...`, the classic `ghp_...` regex doesn't match the newer prefix), Slack incoming webhook URLs, Google API keys (`AIza...`), Stripe live/test secret keys (`sk_live_...`/`rk_test_...`), npm tokens (`npm_...`), and SendGrid keys (`SG....`).
- Guard the MCP stdio server startup behind `require.main === module` and export the pure redaction functions (`redactValue`, `auditExchange`) so they're unit-testable without spinning up the stdio transport — this server previously had zero test coverage despite doing security-critical redaction.
- Add `server.test.js` (Node's built-in `node:test`/`assert`, zero new deps, consistent with the server's "pure Node, zero deps" design) — 12 cases covering every existing and new pattern individually plus an end-to-end `auditExchange` regression check that a JSON-body password and an `Authorization`/`Cookie` header are both redacted from a single captured exchange.
- Updated the `http-evidence` skill doc to describe the widened coverage so agents relying on it have an accurate picture of what's actually redacted.

Scope kept intentionally small: this is a targeted pattern-list + test-coverage fix to one existing MCP server, not a rewrite. No new dependencies, no new servers.

## Verification

```
$ node --test .codex/mcp-servers/http-audit/server.test.js
# tests 12
# pass 12
# fail 0

$ printf '{"jsonrpc":"2.0","id":1,"method":"tools/list"}\n' | node .codex/mcp-servers/http-audit/server.js
# still starts and lists the http_audit tool correctly (require.main guard doesn't change runtime behavior)

$ npx prettier --check .codex/mcp-servers/http-audit/server.js .codex/mcp-servers/http-audit/server.test.js
# clean
```

## Test plan

- [x] `node --test .codex/mcp-servers/http-audit/server.test.js` passes (12/12)
- [x] Server still starts and responds to `tools/list` under the actual MCP stdio transport (unchanged behavior for the real entry point)
- [x] `prettier --check` clean on changed files
- [ ] Reviewer: confirm the new JSON-body regex doesn't over-redact adjacent non-secret JSON fields (see `server.test.js` JSON-body test case for the exact before/after)

---

## Scan summary (part 2 of this run's routine)

Also ran the authorized discovery-only scan against `https://vulnerable-bounty-site.vercel.app` per this run's instructions. That work isn't tied to a code change, so its findings are reported in the run's final chat summary rather than here.

https://claude.ai/code/session_01PehwXFCjgEsyjhNAuh2o8d

---
_Generated by [Claude Code](https://claude.ai/code/session_01PehwXFCjgEsyjhNAuh2o8d)_

### DIFF
--- a/.codex/mcp-servers/http-audit/server.js
+++ b/.codex/mcp-servers/http-audit/server.js
@@ -33,7 +33,16 @@ const SENSITIVE_HEADERS = new Set([
 const SECRET_VALUE_PATTERNS = [
   [/\bAKIA[0-9A-Z]{16}\b/g, "[REDACTED_AWS_KEY]"],
   [/\bgh[pousr]_[A-Za-z0-9]{20,}\b/g, "[REDACTED_GH_TOKEN]"],
+  [/\bgithub_pat_[A-Za-z0-9_]{20,}\b/g, "[REDACTED_GH_TOKEN]"],
   [/\bxox[baprs]-[A-Za-z0-9-]{10,}\b/g, "[REDACTED_SLACK_TOKEN]"],
+  [
+    /\bhttps:\/\/hooks\.slack\.com\/services\/T[A-Za-z0-9]{6,}\/B[A-Za-z0-9]{6,}\/[A-Za-z0-9]{16,}\b/g,
+    "[REDACTED_SLACK_WEBHOOK]",
+  ],
+  [/\bAIza[0-9A-Za-z_-]{35}\b/g, "[REDACTED_GOOGLE_API_KEY]"],
+  [/\b(?:sk|rk)_(?:live|test)_[0-9A-Za-z]{16,}\b/g, "[REDACTED_STRIPE_KEY]"],
+  [/\bnpm_[A-Za-z0-9]{36}\b/g, "[REDACTED_NPM_TOKEN]"],
+  [/\bSG\.[A-Za-z0-9_-]{22}\.[A-Za-z0-9_-]{43}\b/g, "[REDACTED_SENDGRID_KEY]"],
   [
     /\bey[A-Za-z0-9_-]{10,}\.[A-Za-z0-9_-]{10,}\.[A-Za-z0-9_-]{10,}\b/g,
     "[REDACTED_JWT]",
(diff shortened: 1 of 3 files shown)


deonmenezes:
The required `blocking-ci` → `cargo-deny` check is failing (`advisories FAILED`), but it's unrelated to this PR's diff and pre-existing repo-wide:

- **Advisory:** [RUSTSEC-2026-0204](https://rustsec.org/advisories/RUSTSEC-2026-0204) — invalid pointer dereference in `crossbeam-epoch`'s `fmt::Pointer` impl.
- **Where:** `crossbeam-epoch v0.9.18`, pinned in `codex-rs/Cargo.lock` (transitive: `crossbeam-deque` → `ignore` → `codex-file-search` → most of the workspace). That pin predates this PR — this branch never touched `Cargo.lock`/`Cargo.toml`.
- **Why it's showing up here:** `cargo-deny` is a `workflow_call` reusable workflow invoked by `blocking-ci`, so it only runs in a PR context — there's no run against bare `main` to compare against, but since the lockfile entry is unchanged from `main`, this would fail identically on any other PR touching `codex-rs` right now.
- **Suggested fix:** the advisory's own solution is `cargo update -p crossbeam-epoch` (>=0.9.20 fixes it, and 0.9.18→0.9.20 is a semver-compatible patch bump, so no other coordination should be needed).
- **Why I didn't push that fix myself:** I tried `cargo update -p crossbeam-epoch --precise 0.9.20` in this session and it failed — not on the crossbeam bump itself, but because resolving the workspace lockfile at all requires re-fetching `codex-realtime-webrtc`'s git dependency (`juberti-oai/rust-sdks`) and its `libyuv` submodule from `chromium.googlesource.com`, and this session's network egress policy returns 403 on that host. I don't want to hand-edit `Cargo.lock` directly (risks bad checksums / an inconsistent lockfile) or bypass the check, so I'm flagging it instead of forcing something risky.

Recommend fixing this in a separate PR from an environment with full network access (`cd codex-rs && cargo update -p crossbeam-epoch`), rather than folding an unrelated dependency bump into this one. Happy to do it myself if a future run has broader egress access.

---

PR #145, opened 2026-07-17
feat: wire a passive http_recon MCP server for live-target discovery

deonmenezes:
## What gap this closes

`mantis_http_audit` (`.codex/mcp-servers/http-audit`) packages an *already-captured* HTTP exchange into redacted evidence — by design it "makes no network call itself." That left a real hole: the harness had no way to check a **live, black-box target** (a deployed URL with no source to read) during Recon/Detect. `MANTIS.md`'s own roadmap flagged this as a P2 gap ("Recon/DAST toolchain … `httpx`, … each report-until-installed").

This PR wires the passive tier of that gap without any external binary: `mantis_http_recon` performs one bounded `GET` against a URL and reports Detect-stage candidates for:

- missing hardening headers (`Strict-Transport-Security`, `Content-Security-Policy`, `X-Frame-Options`, `X-Content-Type-Options`, `Referrer-Policy`, `Permissions-Policy`)
- `Set-Cookie` flags missing `Secure`/`HttpOnly`/`SameSite`
- `Server`/`X-Powered-By` version disclosure

## Why it's safe to run generously

- **GET only, single request, no payloads.** It is not an injection/auth confirmer — those classes stay gated behind explicit exploit authorization per `mantis-pipeline`.
- **SSRF-guarded.** Refuses to resolve to loopback, link-local (including the `169.254.169.254` cloud-metadata address), or RFC1918/ULA ranges — checked on the initial URL *and every redirect hop*, so a target's own redirect chain can't walk the tool into internal infrastructure.
- **Bounded.** Redirects capped at 3, response body capped at 64KB read / 2KB preview, request timeout capped at 20s.
- **Body preview is redacted** of shape-matched secrets (AWS keys, GH tokens, JWTs, private keys) before it's ever returned.
- Explicitly does **not** attribute this tool to the `recon` agent role, which is contractually read-only/code-only ("do not touch the network or any running system") — `mantis-pipeline` now calls out that a black-box live target is a separate path from that agent.

## Changes

- `.codex/mcp-servers/http-recon/server.js` — new pure-Node MCP server (zero deps), same `mcp_stdio.js` pattern as the other scanners.
- `.codex/config.toml` — registers `mcp_servers.mantis_http_recon`.
- `.codex/skills/http-recon/SKILL.md` — new skill: when to use it, what it won't do, and the canary-tripwire reminder to treat response content as untrusted data.
- `.codex/skills/mantis-pipeline/SKILL.md` — stage 1 now notes the black-box live-target path.
- `.codex/MANTIS.md` — capability table, skill list, and roadmap updated; also fixed a pre-existing prose bug in the roadmap section where a mid-sentence `+` at a line wrap was being reflowed by prettier into a spurious markdown list item, garbling the sentence.

## Testing

Manually smoke-tested the MCP server directly over stdio (`initialize` / `tools/list` / `tools/call`):
- Happy path against a real public URL — correct status/headers/findings shape.
- SSRF guard: refused `169.254.169.254`, `127.0.0.1`, and IPv6 `::1` (also fixed a bug where `new URL().hostname` keeps `[...]` brackets around IPv6 literals, which broke both the DNS-lookup guard and the actual connect — now stripped consistently).
- Invalid URL and non-http(s) scheme (`ftp://`) both rejected with clear errors.
- `npx prettier --check` passes on all changed files.

No existing tests cover `.codex/mcp-servers/*` (none of the sibling scanners have JS unit tests either), so this follows the existing manual-smoke-test convention for this layer.

---
_Generated by [Claude Code](https://claude.ai/code/session_0118qv1zQeKAaYtriQ98Vztz)_

### DIFF
--- a/.codex/MANTIS.md
+++ b/.codex/MANTIS.md
@@ -6,7 +6,7 @@ canonical map of the `.codex/` layer; keep it in sync when you add capability.
 
 The invariant that governs everything (PRD Appendix D):
 
-- **Tool = code** — an MCP server handler that *does* something. Wired in `.codex/config.toml [mcp_servers.*]`, implemented under `.codex/mcp-servers/<name>/server.js`.
+- **Tool = code** — an MCP server handler that _does_ something. Wired in `.codex/config.toml [mcp_servers.*]`, implemented under `.codex/mcp-servers/<name>/server.js`.
 - **Agent = prompt** — a system prompt + model tier + tool/permission policy. A TOML role file under `.codex/agents/<name>.toml`, auto-discovered and offered as a `spawn_agent` `agent_type`.
 - **Skill = knowledge** — reference text the model reads. `.codex/skills/<name>/SKILL.md`.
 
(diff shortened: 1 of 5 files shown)


deonmenezes:
The `cargo-deny` failure (`advisories FAILED`) is `RUSTSEC-2026-0204` against the transitive `crossbeam-epoch 0.9.18` pin in `codex-rs/Cargo.lock` — unrelated to this PR, which only touches `.codex/*` (JS/TOML/Markdown), no Rust code or lockfile. Confirmed pre-existing: `blocking-ci` on `main` at `c3a2e68` (this PR's base) already fails the same way, since cargo-deny checks the live advisory database rather than a pinned snapshot. Bumping/patching that dependency graph is out of scope for this change — flagging rather than pulling it into this PR.

---

deonmenezes:
Second blocker for `CI required`, also pre-existing and unrelated to this diff: `sdk / sdks` timed out at exactly 30 minutes (its configured `timeout-minutes: 30`) on both the original run and the re-run, always at the `Build codex with Cargo` step. Per that job's own comment in `.github/workflows/sdk.yml`, a cold release build of `codex-rs` "takes well over an hour with no cache at all" and relies on an `actions/cache` keyed by `hashFiles('codex-rs/Cargo.lock')`. This PR doesn't touch `Cargo.lock` or any Rust code, so it's using the same cache key as `main` — the timeout means that cache is currently cold/missing for everyone, not something introduced here. Re-running didn't help (same result both times), so I won't keep re-triggering it; it needs either a warm cache (e.g. a successful run on `main` to seed it) or a longer timeout, both out of scope for this change.

Net: `repo-checks`, `codespell`, `rust-ci`, `v8-canary`, and `Blob size policy` all pass on this diff. The two remaining `CI required` blockers (`cargo-deny` advisory, `sdk/sdks` cache/timeout) are pre-existing and repo-wide — flagging for visibility rather than pulling either fix into this PR.

---

PR #148, opened 2026-07-18
fix: close secret-DLP coverage gap between findings guard and http-audit redactor

deonmenezes:
## What gap this closes

`.codex/mcp-servers/findings/server.js` and `.codex/mcp-servers/http-audit/server.js` each kept their own independent copy of the "does this look like a raw secret" pattern list. They had already drifted:

- `findings/server.js`'s `scanForSecrets()` — the backstop that refuses any `finding_create`/`finding_update` payload shaped like a raw credential (PRD section 9/11 secrets/DLP) — was missing two shapes that `http-audit`'s evidence-pack redactor already handled: `user:pass@` URL userinfo, and generically-named `token=`/`password=`/`secret=`-style query/form params.
- Net effect: a finding payload carrying a leaked credential in one of those two shapes would sail past the "never raw secrets" refusal in the findings spine, even though the same credential would have been redacted if it had gone through `http_audit` first. That's a real detection/DLP inconsistency, not just style debt.

## What changed

- Added `.codex/mcp-servers/lib/secret_patterns.js` as the single source of truth for secret-shape patterns, exporting:
  - `detectPatterns()` — fresh non-global `RegExp`s per call, safe for repeated `.test()` (avoids the classic global-regex `lastIndex` statefulness bug)
  - `globalPatterns()` — fresh global `RegExp`s per call, for `.replace()`-based redaction
  - `containsSecret(value)` — the boolean guard
- `findings/server.js` now uses `containsSecret` from the shared module instead of its own pattern list.
- `http-audit/server.js` now builds its redaction table from the shared module's patterns + per-pattern `replacement` instead of its own pattern list.
- Extended the shared list with a few more common secret shapes both guards were missing entirely: Google API keys (`AIza...`), Stripe live keys (`sk_live_`/`rk_live_`), npm tokens (`npm_...`), SendGrid keys (`SG.x.y`).

Net: the two guards can no longer independently drift apart, and both now catch a broader set of real-world secret shapes.

## Why

The Mantis findings spine's stated invariant is "never raw secrets/tokens/cookies/full bodies" land in a finding (PRD section 9/11). Two independently-maintained pattern lists enforcing that same invariant is exactly how gaps like this open up over time — the fix is structural (one shared source of truth), not just adding more patterns.

## Testing

No JS test harness exists yet for the `.codex/mcp-servers/*` servers, so I exercised the real code paths directly:

- `node -c` syntax-checked all three changed/added files.
- `prettier --check` passes on all three files (repo's `.prettierrc.toml`).
- Wrote a standalone Node script exercising `containsSecret()` against 13 cases (7 previously-covered secret shapes, 4 newly-added shapes, 2 benign strings) — all pass.
- Drove both MCP servers end-to-end over their real stdio JSON-RPC protocol (not mocked):
  - `findings/server.js`: a `finding_create` call containing a Google API key is now correctly **refused**; a normal XSS finding is still accepted and persisted.
  - `http-audit/server.js`: an HTTP request with `?token=...` and an `Authorization: Bearer sk_live_...` header comes back with both redacted in the evidence pack.
- Confirmed `.codex/findings/` is gitignored and removed the scratch event log created by the manual server test before committing.

## Scan findings (authorized target)

Not tied to this PR — see the run's final summary / `reports/` for the discovery-only scan against the authorized target for this run.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CeRoFRZXnxWYNaT6SXmgi1)_

### DIFF
--- a/.codex/mcp-servers/findings/server.js
+++ b/.codex/mcp-servers/findings/server.js
@@ -5,6 +5,7 @@ const fs = require("node:fs");
 const path = require("node:path");
 const crypto = require("node:crypto");
 const { createServer } = require("../lib/mcp_stdio.js");
+const { containsSecret } = require("../lib/secret_patterns.js");
 
 /**
  * Mantis findings service -- the tool-owned findings spine (PRD section 9,
@@ -37,17 +38,6 @@ const STATUS_TRANSITIONS = {
 const VALID_STATUSES = Object.keys(STATUS_TRANSITIONS);
 const VALID_SEVERITIES = ["info", "low", "medium", "high", "critical"];
 
-// Cheap secret-shaped-string guard so raw credentials never land in a finding
-// or evidence blob (PRD section 9 "never raw secrets/tokens/cookies/full
-// bodies", section 11 secrets/DLP). This is a backstop, not a full DLP engine.
-const SECRET_PATTERNS = [
-  /\bAKIA[0-9A-Z]{16}\b/, // AWS access key id
-  /\bgh[pousr]_[A-Za-z0-9]{20,}\b/, // GitHub tokens
-  /\bxox[baprs]-[A-Za-z0-9-]{10,}\b/, // Slack tokens
-  /-----BEGIN [A-Z ]*PRIVATE KEY-----/, // PEM private keys
-  /\bey[A-Za-z0-9_-]{10,}\.[A-Za-z0-9_-]{10,}\.[A-Za-z0-9_-]{10,}\b/, // JWTs
-];
-
 function ensureDataDir() {
   fs.mkdirSync(DATA_DIR, { recursive: true });
 }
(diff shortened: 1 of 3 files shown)


deonmenezes:
The `cargo-deny` failure here is pre-existing and unrelated to this PR's diff (which only touches `.codex/mcp-servers/*.js`).

Root cause: `advisories` fails on `RUSTSEC-2026-0204` (invalid pointer dereference in `crossbeam-epoch`'s `fmt::Display`/`fmt::Pointer` impls), pinned via `crossbeam-epoch v0.9.18` in `codex-rs/Cargo.lock`. That pin is identical on `main` — this PR doesn't touch `Cargo.lock` — so `main` will fail the same `cargo-deny` gate the next time its advisory DB is refreshed; this isn't something introduced by this change.

I didn't attempt a fix in this session: `crossbeam-epoch` needs `>=0.9.20` (`cargo update -p crossbeam-epoch`), but `codex-rs` is a large workspace with several git-sourced dependencies (e.g. `libwebrtc`/`rules_rust`/`ratatui` forks) that this sandboxed session's network policy blocks, so I can't run `cargo update` + a full build/test cycle here to verify a bump is safe. That fix belongs in its own dependency-bump PR with full network access rather than bundled into this one.

---

PR #151, opened 2026-09-11
detect: add prototype pollution (CWE-1321) sink coverage to source_sink_scan

deonmenezes:
## What gap this closes

`source_sink_scan`'s heuristic rule set (`.codex/mcp-servers/program-analysis/source_sink_rules.js`) had no coverage at all for **JS prototype pollution (CWE-1321)** — a well-established, high-impact vulnerability class (gadget chains in several Node frameworks have turned prototype pollution into RCE) that generic SAST rules and the other capability servers in the toolchain (semgrep/CodeQL/osv/trufflehog/bandit/trivy) don't reliably flag as a single first-class rule in this pipeline.

Two new sink rules, both tagged `CWE-1321`:

- `js.prototype_pollution.merge` — attacker-controlled data flowing into unguarded merge/extend calls: `_.merge`/`_.mergeWith`/`_.defaultsDeep`/`_.assignIn`, `deepmerge`/`deepmerge.all`, `$.extend`/`jQuery.extend`, `Object.assign`.
- `js.prototype_pollution.dunder_proto` — direct `__proto__` writes, both bracket form (`obj["__proto__"] = ...`) and dot form (`obj.__proto__.x = ...` / `.__proto__[...]`).

These pair naturally with the existing `js.http.query` source rule (`req.query|body|params|headers|cookies`) already in the file, so a request body merged straight into an object now surfaces as a candidate the same way the existing XSS/command-injection/deserialization sinks do.

## Why this is safe and well-scoped

- Read-only, static, additive-only change to one heuristic rule table — no behavior change to any other rule, no change to the scan pipeline, findings lifecycle, or any live-target logic.
- This scanner is explicitly documented as a **recall-oriented heuristic proxy**, not a taint engine (see the file's header comment and the `detector` agent's "do not self-censor" discipline) — new candidates still have to survive reachability tracing and attacker-simulated validation before becoming `confirmed`, so this can only add candidates, never silently suppress anything.
- Checked against the current open-PR backlog for detection work; no open PR already proposes prototype-pollution coverage, so this doesn't duplicate in-flight work.

## Testing

- `node -e` sanity checks against representative vulnerable/safe snippets for both new patterns (merge sinks: `Object.assign`, `_.merge`, `deepmerge.all`, `$.extend`, `jQuery.extend`; dunder-proto: bracket and dot-form writes) — all match as expected, with no false positives on safe reads like `obj.__proto__` alone (no assignment) or `Object.keys(x)`.
[...]

PR #136, opened 2026-07-14
docs: log this run's blocked scan attempt against vulnerable-bounty-site

deonmenezes:
## Context

Part of the recurring 4h maintenance+testing routine, which is authorized to run a discovery-only scan against `https://vulnerable-bounty-site.vercel.app`.

## What happened this run

Same result as the last two runs (see the two prior open scan-log PRs): no scan was possible. Every outbound HTTPS request from this session's environment was rejected at the network egress proxy before any TLS handshake with the origin completed:

- `curl -sS -m 15 https://vulnerable-bounty-site.vercel.app` → `curl: (56) CONNECT tunnel failed, response 403`
- `curl` against an unrelated control domain (`example.com`) → identical `CONNECT tunnel failed, response 403`
- `WebFetch` against the target → `HTTP 403 Forbidden`, rejected before reaching the origin
- `$HTTPS_PROXY/__agentproxy/status` → fresh `connect_rejected` entries in `recentRelayFailures` for `vulnerable-bounty-site.vercel.app:443`, timestamped this run

This confirms (again) it's a categorical "no general internet egress" policy on the environment, not a target-side response and not specific to one request tool. No request has reached the target across three runs now, so there are no scan findings to report.

## What this PR adds

An entry appended to `reports/scans/vulnerable-bounty-site.md` (introduced by an earlier, still-unmerged PR) documenting this run's repeat of the block, with the evidence above. The file is a running log so once the scan can actually reach the target, future entries are diffable against a "was blocked" baseline instead of starting cold.

## Action needed from a human

Add `vulnerable-bounty-site.vercel.app` (or general internet egress) to this environment's outbound network allowlist. This is now the third consecutive run with the identical result — the policy fix, not another retry, is what unblocks this routine.

## Scope

Docs-only addition (single file under `reports/`), no code changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_019F9XMrN6RuRxa6PDP5EHTN)_

PR #152, opened 2026-09-11
fix: close CWE-89 SQLi coverage gap in source/sink heuristic for JS/Python/Go

deonmenezes:
## Summary

`source_sink_scan` (`.codex/mcp-servers/program-analysis/source_sink_rules.js`) is the fast, dependency-free heuristic that tags attacker-controlled sources and dangerous sinks during the Recon/Detect stage of the Mantis pipeline. Before this change, SQL injection (CWE-89) sinks were only tagged for **Java** (`java.statement.execute`), even though every other common sink class (CWE-78 command injection, CWE-79 XSS, CWE-95 code injection, CWE-502 deserialization) is covered across JS, Python, Go, and Java. That left Node, Python, and Go backends — the most common web-app stacks in this scanner's target population — with **zero SQLi sink candidates**, which is a straight detection-recall gap for one of the highest-impact, most commonly requested vulnerability classes (it's explicitly one of the classes this tool is asked to scan for).

## What changed

Added CWE-89 sink rules for JS, Python, and Go, keyed on the actually-risky shape (a query/execute call built from a template literal or string concatenation) rather than the bare call name, so safe parameterized calls don't get flagged and drown out real candidates in noise:

- **JS**: `.query()` / `.execute()` called with a template literal containing `${...}` or with string concatenation (`js.sql.query_template_literal`, `js.sql.query_string_concat`)
- **Python**: `cursor.execute()` called with an f-string, `%`-formatted string, or concatenation (`py.sql.execute_fstring`, `py.sql.execute_percent_format`, `py.sql.execute_string_concat`)
- **Go**: `Query` / `QueryRow` / `QueryContext` / `Exec` called with `fmt.Sprintf(...)` or string concatenation (`go.sql.query_sprintf`, `go.sql.query_string_concat`)

Each new rule is scoped to `kind: "sink"`, `cwe: "CWE-89"`, matching the existing schema consumed by `source_sink_scan` in `server.js` — no changes needed there.

## Why this shape (false-positive control)

A naive `.query(` / `cursor.execute(` match would fire on every DB call, including the safe parameterized form (`db.query(sql, [id])`, `cursor.execute(sql, (id,))`), which is the *recommended* pattern and would dominate results with noise. Instead each rule requires the query text itself to be dynamically built (template interpolation, `%`/f-string formatting, or `+`/`Sprintf` concatenation) — the actual precondition for SQL injection — before tagging it as a candidate sink.

## Testing

Added `.codex/mcp-servers/program-analysis/source_sink_rules.test.js` using Node's built-in `node:test` runner (zero new deps, matching this directory's "pure Node" style). 11 cases covering:
- True positives for each new rule (JS template literal / concat, Python f-string / `%`-format / concat, Go `Sprintf` / concat)
- True negatives confirming parameterized/safe calls in each language are **not** flagged
- A regression check that the existing Java `statement.execute` SQLi rule is unaffected

```
$ node --test .codex/mcp-servers/program-analysis/source_sink_rules.test.js
# tests 11
# pass 11
# fail 0
```

Also ran `prettier --check` on both changed files (clean).

## Scope

Small and focused: one rules file + one new test file, no changes to `server.js`, no new dependencies, no changes to the MCP tool schema or the findings lifecycle.

---

https://claude.ai/code/session_0134vdLGmjsWH1mN7WDkHQq5

---
_Generated by [Claude Code](https://claude.ai/code/session_0134vdLGmjsWH1mN7WDkHQq5)_


deonmenezes:
CI on this PR shows two red checks in `blocking-ci`: `cargo-deny / cargo-deny` and `repo-checks / build-test`. Neither is caused by this diff (one file, `source_sink_rules.js` + a new `node:test` file, no Rust or npm-release changes):

- **`repo-checks / build-test`** fails at the "Stage npm package" step: `scripts/stage_npm_packages.py` tries to download native binaries from a hardcoded upstream workflow run (`https://github.com/openai/codex/actions/runs/26201494185`) and gets `no valid artifacts found to download` — that run's artifacts have expired/aren't accessible from this fork. Nothing in this PR touches packaging.
- **`cargo-deny / cargo-deny`** fails on `advisories FAILED` — a RUSTSEC advisory somewhere in the existing (large) Rust workspace dependency tree. Nothing in this PR touches any `Cargo.toml`/`Cargo.lock` or Rust code.

Both failures match the same pattern already logged on unrelated prior PRs from this same recurring routine (e.g. #151, #136 — different file changes, same two red checks), so this is a pre-existing, base-repo infrastructure issue rather than something introduced here. No fix for either currently exists in the repo to port in, and resolving them (pinning a working native-artifact source for the npm release step; auditing/patching the flagged Rust dependency) is out of scope for a one-file detection-rule change. Not re-running the jobs since both failures are deterministic (a missing external artifact, a real advisory), not flaky.

---

PR #153, opened 2026-09-11
fix(mantis-http-audit): redact secret-bearing fields in JSON bodies

deonmenezes:
## What gap this closes

`mantis_http_audit` (`.codex/mcp-servers/http-audit/server.js`) is the evidence-pack tool the Validate/DAST stage uses to attach reproducible HTTP evidence to a finding without leaking secrets (PRD invariant: "never raw secrets/tokens/cookies/full bodies" in findings/evidence).

Its `SECRET_VALUE_PATTERNS` list already redacted generically-named secret fields (`password`, `token`, `api_key`, `session_id`, etc.) when they appeared as `key=value` in a query string or form-encoded body. It had **no equivalent pattern for the same fields in a JSON body** (`"key": "value"`), which is how the large majority of modern web/API traffic actually carries these fields. A captured login request like:

```json
{"email":"a@example.com","password":"hunter2"}
```

would pass straight through `redactValue`/`summarizeBody` with the plaintext password intact in the evidence pack's body preview — a direct DLP failure for a tool whose entire purpose is to prevent exactly that.

## What changed

- Added a case-insensitive regex to `SECRET_VALUE_PATTERNS` that matches the same secret-bearing field names in JSON `"key":"value"` shape and redacts the value, leaving non-secret fields untouched.
- Guarded the `createServer(...)` stdio-transport startup behind `require.main === module` so the pure redaction/audit functions can be `require`'d and unit-tested without spinning up the MCP stdio listener (no behavior change when run as the actual MCP server).
- Added `.codex/mcp-servers/http-audit/server.test.js` (Node's built-in `node:test`, zero new deps) covering: header redaction, form/query redaction, the new JSON-body redaction (including case-insensitivity and an end-to-end `auditExchange` check that the preview never contains the plaintext secret), existing shape-based redaction (AWS key, JWT), stable request-ref hashing, and the input-validation error path. This is the first test coverage this server had.

## Testing

```
$ node --test .codex/mcp-servers/http-audit/server.test.js
# tests 8
# pass 8
# fail 0

$ npx prettier --check .codex/mcp-servers/http-audit/server.js .codex/mcp-servers/http-audit/server.test.js
```

---

Part of the recurring detection-improvement + authorized-scan maintenance routine on this repo.

https://claude.ai/code/session_01GUFZjUCaiuieUXFSiexQZA

---
_Generated by [Claude Code](https://claude.ai/code/session_01GUFZjUCaiuieUXFSiexQZA)_


deonmenezes:
Closing this as a duplicate. While chasing down the `repo-checks / build-test` CI failure on this PR, I found it's unrelated to this diff (see below) — but in the process I also found this exact fix already sitting in two open PRs from earlier runs of this same recurring routine:

- #143 "fix: widen mantis_http_audit secret redaction to close JSON-body + provider-key gaps" (2026-07-17) — same JSON-body regex, same `require.main === module` guard, same test file, plus several provider-key patterns (Google, Stripe, npm, SendGrid, Slack webhooks, GH fine-grained PATs) this PR didn't add. This is the more complete version of this fix.
- #148 "fix: close secret-DLP coverage gap between findings guard and http-audit redactor" (2026-07-18) — a structurally different but overlapping fix that extracts a shared `lib/secret_patterns.js` so `findings/server.js` and `http-audit/server.js` can't drift apart, and adds several of the same provider-key patterns (though not the JSON-body regex itself).
- #130 (2026-07-13) was apparently the first attempt at this same gap.

Recommend reviewing/merging #143 (most complete for this specific gap) and separately considering #148's shared-module refactor, rather than this PR.

Unrelated finding worth flagging: the `repo-checks / build-test` failure on this PR's head commit is `stage_npm_packages.py` failing to download native artifacts from a hardcoded upstream reference (`openai/codex` run `26201494185`, pinned in the initial fork commit `d155094`) — "no valid artifacts found to download," almost certainly because that run's artifacts expired (GitHub Actions artifacts expire after 90 days by default, and that pin hasn't been touched since). This is unrelated to any of this PR's changes and would fail identically on `main` and on every other open PR right now.

---

PR #154, opened 2026-09-12
fix: close SQL-injection coverage gap in source_sink_scan heuristic

deonmenezes:
## What gap this closes

`mantis_program_analysis`'s `source_sink_scan` (`.codex/mcp-servers/program-analysis/source_sink_rules.js`) is the dependency-free regex proxy for attacker-controlled sources and dangerous sinks used in Recon/Detect (FR-3.1/3.2). In this environment none of the external scanner binaries (`semgrep`, `codeql`, `bandit`, `trivy`, `osv-scanner`, `trufflehog`, `z3`, `ast-grep`) are installed, so this heuristic is currently the **only** static detector that actually runs.

It had **zero SQL-injection sinks** for JS/TS, Python, or Go, and only a blanket `Statement.execute()` rule for Java. SQLi is one of the most common and highest-impact vulnerability classes (CWE-89) and was entirely unrecognized across 3 of the 4 supported languages.

## What changed

Added precise sinks (tagged `CWE-89`) for the unsafe query-construction forms, scoped narrowly to avoid flagging parameterized queries:

- **JS/TS**: `.query()`/`.execute()` called with a template literal containing `${...}` interpolation, or with string concatenation (`"..." + x`).
- **Python**: `.execute()` called with an f-string, or with string concatenation/`%`-formatting.
- **Go**: `.Query()`/`.QueryRow()`/`.Exec()` called with `fmt.Sprintf(...)`, or with string concatenation.

Also updated `.codex/skills/program-analysis/SKILL.md`'s example sink list to mention the new SQLi coverage.

## Why these patterns keep false positives low

Each pattern only matches the *unsafe construction* form (interpolation/concatenation/formatting built directly into the query string), not the safe parameterized form (`query("... ?", [id])` / `execute("... %s", (id,))` / `Query("... ?", id)`), which is by far the more common correct usage and would otherwise dominate the output with noise.

## Testing

No existing test harness covers `.codex/mcp-servers/`. Verified manually:

- `prettier --check` passes on both changed files.
- A standalone Node script exercised all 6 new rules against paired vulnerable/safe one-liners (template literal, string concat, f-string, `%`-format, `fmt.Sprintf`) — 13/13 assertions passed, correctly matching the vulnerable form and correctly skipping the parameterized-safe form in each case.
- Ran the heuristic end-to-end (mirroring `sourceSinkScan`'s file-walk logic) against a small synthetic fixture with one vulnerable + one safe query per language — output correctly flagged only the vulnerable lines.

https://claude.ai/code/session_01MVEi1xrGAB7YRCy4MeG5br

---
_Generated by [Claude Code](https://claude.ai/code/session_01MVEi1xrGAB7YRCy4MeG5br)_


deonmenezes:
Closing as a duplicate: #152 (opened by an earlier run of this same routine, 2026-09-11) already covers this exact gap — CWE-89 sinks for JS/`.query()`/`.execute()`, Python `cursor.execute()`, and Go `Query`/`QueryRow`/`Exec`, with the same false-positive-avoidance approach, plus a proper `node:test` test file that this PR didn't have. No need to carry two copies of the same fix.

Broader note for whoever triages this: `repo-checks / build-test` (via `blocking-ci`) has failed on every PR from this recurring routine going back to at least run #59 (2026-07-17) — a `stage_npm_packages.py` step hardcodes a native-artifact download from a specific `openai/codex` Actions run (`26201494185`) that has since expired ("no valid artifacts found to download"). This is unrelated to any of these PRs' content, but it means none of them can go green, so nothing from this routine has merged in ~2 months and duplicate PRs for the same gap (SQLi alone: #104, #101, #114, #127, #139, #142, #152, this one) keep piling up because each run can't see prior work as landed. Worth fixing the artifact-source step (or exempting non-release PRs from it) and then bulk-triaging the ~50 open draft PRs before this routine keeps running.

---

PR #155, opened 2026-09-12
detect: add Zip Slip (CWE-22 archive extraction) sink coverage

deonmenezes:
## Summary

- Closes a detection coverage gap in `source_sink_scan` (`.codex/mcp-servers/program-analysis/source_sink_rules.js`): the heuristic scanner had rules for direct path sinks (`fs.readFile`, `send_file`, `http.ServeFile`, ...) but nothing for the **Zip Slip** class (CWE-22), where the attacker-controlled path segment is an *archive entry name* rather than a request parameter. A malicious `../../../etc/cron.d/x`-style entry name lets an unchecked extraction overwrite arbitrary files outside the target directory (the CVE-2018-1000632 pattern that has recurred across many ecosystems).
- Adds one sink rule per language for the classic unsafe-extraction call shape:
  - **JS**: `adm-zip`'s `.extractAllTo(`, the `tar` package's `tar.extract(`
  - **Python**: `zipfile`/`tarfile`'s `.extractall(`, with a negative lookahead that exempts the Python 3.12+ `filter=` guard (mirrors the existing `py.yaml.load_unsafe` safe-loader exemption style already in this file)
  - **Go**: `archive/zip`'s `zip.OpenReader(`
  - **Java**: the `ZipInputStream` read loop's `getNextEntry(`
- All new rules are tagged `cwe: "CWE-22"` and follow the file's existing recall-not-proof heuristic philosophy (surfaces a candidate sink for the Validate stage to trace, doesn't claim reachability itself).
- Updated `.codex/skills/program-analysis/SKILL.md`'s `source_sink_scan` description to mention the new archive-extraction sink examples.

This repo's routine already has a large number of open PRs adding sink coverage for other vulnerability classes (SQLi, XSS/SSTI, SSRF, XXE, path traversal, prototype pollution, NoSQL injection, open redirect, weak crypto, etc.) — I checked the open PR list first and confirmed none of them touch archive/zip extraction, so this is a genuinely new, non-duplicated gap rather than another pass at an already-claimed one.

(Separately: with 50+ open PRs against `main` right now, most several weeks old and none merged, several of them literally duplicating each other's SQLi/secret-redaction coverage — that backlog itself looks worth the repo owner's attention independent of this PR.)

## Test plan

- [x] Added `.codex/mcp-servers/program-analysis/source_sink_rules.test.js` (Node's built-in test runner, matching the pattern used by other open PRs against this file) with one positive + one negative/unrelated-code case per new rule, plus a regression spot-check on two pre-existing rules.
- [x] `node --test .codex/mcp-servers/program-analysis/source_sink_rules.test.js` — 7/7 pass.
- [x] `npx prettier --check` on all three changed files — clean, no formatting drift.

https://claude.ai/code/session_01F1CRFYxx1NoTA5gd2cH62K

---
_Generated by [Claude Code](https://claude.ai/code/session_01F1CRFYxx1NoTA5gd2cH62K)_


deonmenezes:
The failing `Vercel` check (deploying `opentradex-dashboard`) isn't this PR's — this change only touches `.codex/mcp-servers/program-analysis/*` and a skill doc, nothing web-app-shaped. The identical check fails the same way on PR #136, which only adds a line to a markdown log file, so it's not reacting to code content at all. This looks like a stray Vercel project/GitHub integration attached at the account level (project name doesn't match this repo) rather than a check this repo's own CI defines — I don't have access to the Vercel project settings to fix it, and there's no in-repo config for it that I can find. Flagging rather than fixing.

---

deonmenezes:
Two more failing checks on this run, neither caused by this diff (JS-only change under `.codex/mcp-servers/program-analysis/` and a skill doc — no Rust, no npm-packaging files touched):

- **`repo-checks / build-test`**: fails in the `stage_npm_packages.py` step trying to download native build artifacts from a *hardcoded* upstream `openai/codex` Actions run (`WORKFLOW_URL=".../openai/codex/actions/runs/26201494185"`, pinned `CODEX_VERSION=0.133.0-alpha.4`) — `gh run download` returns `no valid artifacts found to download`. GitHub Actions artifacts expire (90-day default retention), so a run pinned by ID will eventually 404 regardless of what any PR here changes. This will fail identically on `main` and on every open PR until that pinned workflow URL/version is refreshed to a current `openai/codex` release — that's an unrelated repo-maintenance task, not something this PR's diff can fix.
- **`cargo-deny / cargo-deny`**: `advisories FAILED` on a real advisory — **RUSTSEC-2026-0274** (double-free/use-after-free in `rtrb::ReadChunk::commit` when a dropped element panics) against `rtrb 0.3.3` pinned in `codex-rs/Cargo.lock:852`. Mitigation per the advisory is bumping to `rtrb` 0.3.5+ (0.3.x line) or newer. This PR doesn't touch any Rust crate or `Cargo.lock`, so it can't have introduced or fixed this; it's a pre-existing dependency-advisory gap in the Rust workspace. I didn't search the open-PR backlog for an existing `rtrb` bump before writing this — if one doesn't exist, that bump looks like a small, well-scoped fix for a follow-up PR, but it's out of scope to fold into this JS-only detection change.

Not re-running either check: both are deterministic (a permanently stale artifact reference, and a pinned vulnerable dependency version), not flakes — a re-run would reproduce the same failure.

---

PR #156, opened 2026-09-12
fix: stop finding_update grade from clobbering previously-graded axes

deonmenezes:
## Summary

`finding_update`'s `grade` handling (`.codex/mcp-servers/findings/server.js`) replaced a finding's whole `grade` object with whatever partial set of axes the caller passed:

```js
if (grade) {
  const disp = gradeToDisposition(grade);
  changes.grade = { ...grade, total: disp.total };
  changes.disposition = disp.disposition;
}
```

`gradeToDisposition` treats any axis missing from the passed `grade` as `0`. So a second `finding_update` call that only re-scores one axis (e.g. a verifier bumping `report_quality` after tightening the write-up, or a later stage correcting `chain` once a fuller exploit chain is traced) silently reset every other previously-graded axis to 0 and recomputed a much lower total/disposition -- with zero grading input on those axes. A finding correctly scored to `SUBMIT` (>=40) could drop to `HOLD`/`SKIP` purely as a side effect of an incremental re-grade, which is exactly the kind of silent false-negative this routine is meant to close (a real finding never reaching the report, not because it stopped scoring well, but because of how the grade was recorded).

I checked all 30 currently-open PRs against this repo first. Three touch `findings/server.js`'s grading/lifecycle logic (#137 enforces per-axis ceilings on a single grade payload, #124 closes a `reasoning_trace_ref` confirm-without-proof loophole, #140 widens the findings-spine secret-DLP backstop) -- none of them touch the cross-call merge behavior this PR fixes; I verified against each of their diffs before starting. This fix also composes cleanly with #137: axis validation and cross-call merging are orthogonal, and I built the test file fresh since `server.test.js` doesn't exist on `main` yet (only on #137's still-unmerged branch).

## What changed

- Merge the incoming `grade` onto the finding's existing `grade` (`{ ...(existing.grade || {}), ...grade }`) before computing the total/disposition, so axes not present in a given call keep their last recorded value; axes that *are* present still overwrite (so a genuine correction still works).
- Updated the `grade` field's tool description and the `findings-spine` skill doc to document that grading is cumulative across calls.
- Added `.codex/mcp-servers/findings/server.test.js` (Node's built-in `node:test`, using the same require-a-patched-copy technique as #137's test file, adapted since it doesn't exist on `main`): a regression test that reproduces the exact clobber (grade all 5 axes, then re-send only one, assert the other four and the total survive), a test that an explicitly re-sent axis does overwrite its own prior value, and a test that a first single-axis grade still computes correctly.

## Testing

```
$ node --test .codex/mcp-servers/findings/server.test.js
# tests 3
# pass 3
# fail 0
```

Confirmed the regression test fails against the pre-fix code (`git stash` back to the original `grade` handling): the merged/second-grade total came back as 15 instead of the expected 85, reproducing the clobber.

```
$ npx prettier --check .codex/mcp-servers/findings/server.js .codex/mcp-servers/findings/server.test.js .codex/skills/findings-spine/SKILL.md
```

## Scope

Small and focused: one behavioral fix in `findings/server.js`, one new test file, one doc sentence. No schema changes, no new dependencies, no changes to any other MCP server.

---

## Part 2 of this run: authorized scan against `vulnerable-bounty-site.vercel.app`

Per the recurring routine, I also attempted the authorized discovery-only scan against `https://vulnerable-bounty-site.vercel.app`. Same result as every prior run logged in the still-open PRs #126/#131/#136: this environment's outbound network policy rejects the connection at the proxy level before any TLS handshake reaches the origin --

```
$ curl -sS -m 15 https://vulnerable-bounty-site.vercel.app
curl: (56) CONNECT tunnel failed, response 403
```

and `$HTTPS_PROXY/__agentproxy/status` shows a fresh `connect_rejected` entry for `vulnerable-bounty-site.vercel.app:443` timestamped this run. No request reached the target, so there are no scan findings to report this run either. I did not open another duplicate scan-log PR for this (three already exist, open and unmerged, saying the same thing) -- flagging it here and in my final summary instead.

https://claude.ai/code/session_01RiwL9XbamAcEqabhTLT3R6

---
_Generated by [Claude Code](https://claude.ai/code/session_01RiwL9XbamAcEqabhTLT3R6)_


deonmenezes:
## CI red: `repo-checks / build-test` ("Stage npm package" step) — not this PR's

The `blocking-ci` / `repo-checks / build-test` check is failing on this PR's head (f115d9a), but the failure is not caused by this diff (which only touches `.codex/mcp-servers/findings/server.js`, its new test file, and one skill doc).

**Root cause:** the "Stage npm package" step (`.github/workflows/repo-checks.yml:57-58`) pins a specific *upstream* `openai/codex` workflow run for its native-binary artifacts:

```yaml
CODEX_VERSION=0.133.0-alpha.4
WORKFLOW_URL="https://github.com/openai/codex/actions/runs/26201494185"
```

That run's artifacts have aged out of GitHub's retention window, so `gh run download` now fails with "no valid artifacts found to download" / exit 1 — before any test in this PR's own files even runs.

**This is not PR-specific.** I checked this fork's last 10 `blocking-ci` runs (`gh`/API, run #62 through #71, spanning 2026-07-18 through today): all 10 fail at this exact step, across completely unrelated diffs — a docs-only scan-log PR (#136), a Zip Slip detection PR (#155), an SQLi coverage PR (#152), an http-audit JSON-redaction PR (#153), a prototype-pollution PR (#151) — none of which touch the npm packaging path. So I'm not spending this PR's one allowed re-run on it: a re-run would reproduce the identical `gh run download` failure since the artifacts aren't coming back, and 10/10 recent runs already rule out a transient flake.

**Likely why the 30-PR backlog never merges:** if `blocking-ci` is a required check, every open PR against this repo has been unable to go green for at least the ~2 months this pinned URL has been stale — independent of how good any individual detection PR is.

**Proposed fix (out of scope for this PR — a pure grading-logic fix shouldn't also carry an unrelated CI/workflow change):** stop pinning to an external, expiring `openai/codex` run. Options: (a) re-pin `WORKFLOW_URL`/`CODEX_VERSION` to a currently-valid upstream run as a stopgap, but that just delays the same failure again once *that* run's artifacts expire; (b) have this fork produce its own native artifacts via a scheduled/on-push run of its own `rust-release`/`mantis-npm-release` workflow (neither has ever run in this fork per the Actions history) and point `stage_npm_packages.py` at that instead, so the source of truth doesn't depend on an external repo's ephemeral retention window.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RiwL9XbamAcEqabhTLT3R6)_

---

deonmenezes:
## Second CI red on this PR's head: `cargo-deny / cargo-deny` — also not this PR's

Same pattern as the `repo-checks / build-test` failure above: `cargo-deny` is also red on this PR's head, and this diff touches zero Rust/Cargo files (only `.codex/mcp-servers/findings/server.js`, its test, and one skill doc), so it can't be the cause.

The job log shows `advisories FAILED, bans ok, licenses ok, sources ok` — i.e. `cargo-deny check advisories` is finding at least one RustSec advisory somewhere in the (very large) `codex-rs` dependency tree that isn't allow-listed in `codex-rs/deny.toml`. I wasn't able to pull the exact advisory id/crate out of the log (the portion I could retrieve was truncated to the dependency-graph context lines, not the leading diagnostic), but I don't need it to establish this isn't PR-specific:

I checked job `cargo-deny / cargo-deny` on run `34673239436` (PR #155, "Zip Slip archive-extraction coverage" — a JS-only change to `.codex/mcp-servers/program-analysis/`, nothing in `codex-rs/`) and it fails identically (`conclusion: failure`). Same check, same failure mode, on a diff with no Rust changes at all -- this is pre-existing on `main`, not something either PR introduced.

Not spending this PR's one re-run on it either, for the same reason as the npm-packaging failure: a RustSec advisory doesn't resolve itself between two runs seconds apart, and the PR #155 comparison already rules out a per-PR flake.

**Net effect:** combined with the npm-packaging failure already documented above, both of this repo's two "blocking" checks (`repo-checks / build-test` and `cargo-deny / cargo-deny`) are red on every PR right now, independent of diff content -- which is consistent with why the 30-PR backlog hasn't been merging.

**Proposed fix (also out of scope for this PR):** run `cargo deny check advisories` locally against `codex-rs/deny.toml` to identify the specific RUSTSEC id(s) now failing, then either bump/patch the offending crate version if a fix is available, or add a scoped, time-boxed `[advisories.ignore]` entry in `deny.toml` citing the advisory id and why it's accepted for now (per that file's existing convention) if no upgrade path exists yet.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RiwL9XbamAcEqabhTLT3R6)_

---

PR #158, opened 2026-09-12
detect: add SSRF (CWE-918) sink coverage to source_sink_scan

deonmenezes:
## What gap this closes

SSRF is explicitly called out in this repo's own pipeline docs as a class the static scanners **don't reliably catch**: `.codex/skills/mantis-pipeline/SKILL.md` lists it among the classes needing "LLM reasoning for classes scanners miss", and `detector.toml`/`recon.toml`/`chain-builder.toml` all mention SSRF as an LLM-only concern. But `source_sink_scan`'s heuristic rule table (`.codex/mcp-servers/program-analysis/source_sink_rules.js`) had **zero sink rules for outbound HTTP client calls** in any of its four supported languages — so a handler that reads a URL from user input and passes it straight into an HTTP client produced no candidate at all from the fast static pass, leaving detection entirely dependent on the model noticing it unaided each run.

Adds one `CWE-918` sink rule per already-supported language, pairing with each language's existing source rules:

- **JS/TS** (`js.ssrf.http_client`): `axios`/`axios.get|post|put|patch|delete|head|request`, `fetch`, `got`, `needle.get|post|put|head`, bare `request(`, `http.request|get`, `https.request|get` — pairs with the existing `js.http.query` source (`req.query|body|params|headers|cookies`).
- **Python** (`py.ssrf.http_client`): `requests.get|post|put|patch|delete|head|request`, `httpx.get|post|put|patch|delete|head|request`, `urllib.request.urlopen`/`urllib2.urlopen` — pairs with `py.flask.request`/`py.django.request`.
- **Go** (`go.ssrf.http_client`): `http.Get|Post|PostForm|Head|NewRequest`, `client.Do(req)` — pairs with `go.http.request`.
- **Java** (`java.ssrf.http_client`): `.openConnection()`/`.openStream()`, `HttpClient.newHttpClient()`, `new HttpGet|HttpPost|HttpPut|HttpDelete(...)` — pairs with `java.servlet.request`.

## Why this is safe and well-scoped

- Read-only, static, additive-only change to one heuristic rule table — no change to any existing rule, the scan pipeline, the findings lifecycle, or any live-target/active-scan logic.
- This tagger is explicitly documented (file header) as a recall-oriented proxy, not a taint engine — new candidates still have to survive reachability tracing and attacker-simulated validation before becoming `confirmed`. This can only add candidates for Validate to triage, never suppress anything.
[...]

PR #159, opened 2026-09-13
detect: derive CVSS v3 severity for OSV records without database_specific.severity

deonmenezes:
## Context

Part of the recurring 4h maintenance+testing routine (review detection logic, make one small focused improvement).

## Gap

`mantis_osv_scanner`'s `osv_scan` only reads severity from `vuln.database_specific.severity`. That field is populated by GHSA-sourced advisories (npm, GitHub Actions, NuGet, ...), but most OSV-native ecosystems -- PyPI's `PYSEC-*`, crates.io's `RUSTSEC-*`, Go's `GO-*` -- never set it. Those advisories instead publish a raw CVSS v3 vector under the OSV schema's top-level `severity` array, which the server ignored entirely.

Net effect on main today: a large fraction of `osv_scan` findings (anything not from a GHSA-backed ecosystem) come back with `severity: "unknown"` even when the advisory carries a real, scoreable CVSS vector -- including critical-severity CVEs. `"unknown"` also isn't a member of `mantis_findings`' `VALID_SEVERITIES`, so an agent that forwards it straight into `finding_create` gets a hard rejection.

## Fix

- Added `.codex/mcp-servers/lib/cvss.js`: a pure, dependency-free CVSS v3.0/v3.1 base-score calculator implementing the FIRST.org spec's base-metric equations (impact, exploitability, the scope-changed branch, and the spec's own `Roundup` rounding), plus the standard qualitative bucket table (none/low/medium/high/critical). Hand-verified against the well-known CVE-2021-44228 (Log4Shell) vector, which scores exactly 10.0/critical.
- `osv_scan` now falls back to this calculator **only when** `database_specific.severity` is absent, so existing GHSA-vocabulary behavior is completely unchanged (no overlap with any severity-vocabulary-mapping work). When the fallback fires, the finding also carries `cvss_vector`/`cvss_score` for transparency. An unparseable or unsupported vector (e.g. `CVSS_V4`, which isn't handled here) still safely falls back to `"unknown"` rather than fabricating a score -- consistent with the repo's "report rather than fabricate" convention for every other scanner server.
- Updated the `osv_scan` tool description and the `osv-dependency-scan` skill doc to explain where `severity` comes from.

## Tests

- `.codex/mcp-servers/lib/cvss.test.js` (9 cases): known-vector regression (Log4Shell = 10.0/critical), a hand-computed medium case with the arithmetic shown in a comment, CVSS 3.0 vs 3.1 acceptance, ignoring trailing temporal/environmental metrics, malformed/unsupported-version/missing-metric inputs returning `null` rather than guessing, and the qualitative bucket boundaries from the spec table.
- `.codex/mcp-servers/osv-scanner/server.test.js` (5 cases, using the same require-patching pattern as the existing `findings/server.test.js` pattern in-flight elsewhere): GHSA severity is preserved unmodified, CVSS fallback fires and returns the right bucket + vector + score, absent-both and malformed-vector cases stay `"unknown"`, and `findCvssV3Vector` correctly ignores non-`CVSS_V3` entries.
- `node --test .codex/mcp-servers/lib/cvss.test.js .codex/mcp-servers/osv-scanner/server.test.js` -- 14/14 passing.
- `npx prettier --check` clean on all touched files.

## Scope

Two files touched by logic changes (`osv-scanner/server.js`, one new `lib/cvss.js` module) plus tests and a one-line skill doc update. Does not touch `source_sink_rules.js`, `http-audit/server.js`, or `findings/server.js`/`canary/server.js`, which already have several other open PRs in flight against them -- kept this change on an untouched surface to avoid adding to that merge-conflict pile.

---

https://claude.ai/code/session_011i5mgPhGzSdejyfeWED8Ec

---
_Generated by [Claude Code](https://claude.ai/code/session_011i5mgPhGzSdejyfeWED8Ec)_


deonmenezes:
Pushed a fix for the one real CI failure this PR caused: Codespell flagged `unparseable` (should be `unparsable`) in `lib/cvss.js`/`lib/cvss.test.js`, now fixed and verified locally with `codespell`.

The other two failing checks are **not caused by this PR**:

- **`repo-checks / build-test`**: fails inside a `gh run download --name x86_64-unknown-linux-musl ... --repo openai/codex 26201494185` step (unpacking a pre-built native artifact from the upstream `openai/codex` repo by a hardcoded workflow-run id) with exit code 1 — nothing in this PR's diff touches Rust code, native builds, or that workflow.
- **`cargo-deny / cargo-deny`**: fails its `advisories` check.

Confirmed both are pre-existing/base-branch issues, not a regression from this diff: PR #158 (`detect: add SSRF (CWE-918) sink coverage to source_sink_scan`, an unrelated diff touching only `source_sink_rules.js`) fails on the exact same two checks (`repo-checks / build-test` and `cargo-deny / cargo-deny`) against the same `main` base. I searched the open-PR queue for an existing fix for either and found none, so there's nothing to port yet — these need attention independent of any single detection PR. Not re-running them since the failures aren't flaky (identical, reproducible root cause on an unrelated PR), so a re-run wouldn't change the outcome.

This PR's own tests (14/14, `node --test .codex/mcp-servers/lib/cvss.test.js .codex/mcp-servers/osv-scanner/server.test.js`) and formatting/spelling checks are clean on the latest commit.

---

PR #161, opened 2026-09-13
Add SQL injection sink rules to source/sink heuristic scanner

deonmenezes:
## What gap this closes

The `source_sink_scan` heuristic (`.codex/mcp-servers/program-analysis/source_sink_rules.js`), which feeds the Detect/Recon stages of the Mantis pipeline (see `.codex/agents/detector.toml`, `.codex/skills/mantis-pipeline`), tagged **CWE-89 (SQL injection) sinks only for Java** (`statement.execute*`). JS, Python, and Go — the languages Mantis is most likely to scan in real web apps — had **zero SQL injection coverage** in this fast heuristic pass, even though SQLi is one of the most common and most pattern-detectable web vulnerability classes. A raw query built via template-literal interpolation, string concatenation, Python `%`-formatting/`.format()`/f-strings passed straight to `execute()`, or `fmt.Sprintf()` fed into `Query`/`Exec` would never surface as a `candidate` for this scanner, meaning Mantis had to rely entirely on semgrep/CodeQL (or LLM reasoning) to catch it in this lane — degrading recall in the fast, dependency-free first pass the PRD calls out as FR-3.1/FR-3.2.

## What changed

Added 8 new sink rules (CWE-89), consistent in style with the existing heuristic rules in the file:

- **JS/TS**: `js.sql.query_template_interp` (template-literal interpolation directly inside `.query(`/`.execute(`), `js.sql.query_concat` (string concatenation into `.query(`/`.execute(`)
- **Python**: `py.sql.execute_fstring` (f-string passed to `cursor/cur/conn/db.execute(`), `py.sql.execute_format_or_percent` (`.format()` or `%`-formatting before `execute(`), `py.sql.execute_concat` (string concatenation into `execute(`)
- **Go**: `go.sql.query_sprintf` (`fmt.Sprintf` fed into `Query`/`QueryRow`/`Exec`), `go.sql.query_concat` (string concatenation into `Query`/`QueryRow`/`Exec`)

The concat rules use a quote-backreference (`` (["'])(?:(?!\1)[\s\S])*\1 ``) rather than a naive `[^"']*` character class, so they still match when the SQL string itself embeds a literal quote of the other kind (e.g. `"... = '" + name`) — a very common real-world SQLi pattern that a naive regex would silently miss.

Parameterized/placeholder calls (e.g. `db.query("... = ?", [id])`, `cursor.execute("... = %s", (id,))`) are deliberately left unmatched, since those are the safe pattern this heuristic should *not* flag — matching the detector's "high recall, but not indiscriminate noise" design intent.

## Verification

- Hand-built safe/unsafe fixtures for each language (unsafe: template interpolation, concatenation incl. embedded-quote case, f-string, `.format()`, `%`-formatting, `fmt.Sprintf`; safe: parameterized calls) — all new rules fire only on the unsafe cases and stay silent on the safe/parameterized ones.
- `node -e "require(...)"` sanity check that the module still parses and exports 34 rules (26 previous + 8 new).
- `npx prettier --check` passes on the changed file.
- `python3 scripts/asciicheck.py` passes on the changed file.

No behavior change to any existing rule; this is purely additive.

---

https://claude.ai/code/session_01WFdnVJsuazt8CkaQe39wQG

---
_Generated by [Claude Code](https://claude.ai/code/session_01WFdnVJsuazt8CkaQe39wQG)_


deonmenezes:
Two `blocking-ci` checks are red on this PR, but neither is caused by this diff (which only touches `.codex/mcp-servers/program-analysis/source_sink_rules.js`):

1. **`repo-checks / build-test`** fails at the "Stage npm package" step: `scripts/stage_npm_packages.py` has a hardcoded `WORKFLOW_URL` pinned to an upstream `openai/codex` Actions run (`26201494185`) to fetch native binary artifacts, and that run's artifacts have expired ("no valid artifacts found to download" for `x86_64-unknown-linux-musl`). This same job succeeded on `main`'s last CI run back on 2026-07-07 (artifacts were presumably still live then), so this is a time-based staleness issue in a pinned external reference, not something introduced by this PR. Fix would be updating `WORKFLOW_URL`/`CODEX_VERSION` in that script to a current `openai/codex` release run — out of scope for this single-file detection-rule change.

2. **`cargo-deny / cargo-deny`** fails on `advisories` (`advisories FAILED, bans ok, licenses ok, sources ok`). This is also pre-existing: the same job failed identically on `main`'s last CI run (2026-07-07, run [28905675542](https://github.com/deonmenezes/mantishack/actions/runs/28905675542)). This PR doesn't touch any `Cargo.toml`/Rust code, so it can't be the cause. Full advisory detail is in the job log: https://github.com/deonmenezes/mantishack/actions/runs/34747851427/job/103698956946

I don't have a fix ready to port for either (both need a maintainer call — bumping the pinned npm-staging workflow reference, and resolving/accepting the specific RUSTSEC advisory), so I'm not widening this PR to cover them. Flagging here rather than staying silent on the red checks. The actual change in this PR (new SQL-injection heuristic rules) is covered by `Codespell`, `Blob size policy`, `sdk / python-sdk`, and the prettier/ASCII checks I ran locally, all of which pass.

---

deonmenezes:
Follow-up: the `CI required` gate is also red, but it's just the aggregator failing because of the two jobs already flagged above, plus a third pre-existing issue: `sdk / sdks` was **cancelled**, not failed — its "Build codex with Cargo" step ran the full 30 minutes (`.github/workflows/sdk.yml` sets `timeout-minutes: 30` on that job) and got killed by the timeout before finishing a from-scratch build of the whole Rust workspace. The same step took the full ~30 minutes on `main`'s last CI run too (2026-07-07, job [85752118988](https://github.com/deonmenezes/mantishack/actions/runs/28905675542/job/85752118988)), so a cold-cache full workspace build being marginal against a 30-minute budget is an existing timing issue, not something this PR's single-file change affects.

So `CI required` isn't a new/independent blocker — it's downstream of the three already-identified pre-existing issues (stale npm-staging artifact URL, cargo-deny advisories, and this build-timeout margin). No fix from this PR to push for any of them.

---

PR #163, opened 2026-09-13
Add SQL injection sinks to the source_sink_scan heuristic

deonmenezes:
## What gap this closes

`source_sink_scan` (`.codex/mcp-servers/program-analysis/source_sink_rules.js`) is the one detector in the capability layer that "works now" with zero external binaries, so it's the first thing that actually runs against a target. Auditing its rule table turned up a real coverage gap: **there was no SQL-injection sink for JS/TS or Python** — only Java's `statement.execute(...)` was tagged `CWE-89`. SQLi is one of the most common web vuln classes this pipeline is meant to catch, and two of the most common web-app languages had zero coverage for it.

## What changed

Added CWE-89 sink rules:

- **JS/TS**: `js.sql.query_template_literal` — `.query`/`.execute`/`.raw` called with a template literal containing `${...}` interpolation (e.g. `` db.query(`SELECT * FROM users WHERE id = ${req.query.id}`) ``); `js.sql.query_string_concat` — same call shapes built via string `+` concatenation.
- **Python**: `py.sql.execute_fstring` — `cursor.execute(f"...")`; `py.sql.execute_string_concat` — string `+` concatenation into `.execute(...)`; `py.sql.execute_percent_interpolation` — `%`-interpolating the string *before* passing it to `.execute(...)` (as opposed to the safe `cursor.execute("... %s", (val,))` form, which is deliberately left unmatched).

Patterns are shaped to stay quiet on the parameterized/safe call shapes (placeholder + separate args array/tuple) rather than firing on every `.query`/`.execute` call, to keep this recall-oriented heuristic from drowning validators in noise.

Also updated `.codex/skills/program-analysis/SKILL.md`'s one-line description of `source_sink_scan`'s sink coverage to mention SQLi, and added `.codex/mcp-servers/program-analysis/source_sink_rules.test.js` (Node's built-in `node:test`, zero new deps, consistent with the "pure Node, zero deps" pattern the rest of the capability layer follows) covering both the vulnerable and safe shapes for each new rule.

## Testing

```
# 4 pass, 0 fail
npx prettier --check .codex/mcp-servers/program-analysis/*.js .codex/skills/program-analysis/SKILL.md
# All matched files use Prettier code style!
```

Also hand-verified end-to-end against small fixture files (a vulnerable Node route and a vulnerable Python handler) to confirm the real `source_sink_scan` handler — not just the raw regexes — reports the new `CWE-89` findings.

---

https://claude.ai/code/session_01DxMc1KFYqNjAwN3xojqRRx

---
_Generated by [Claude Code](https://claude.ai/code/session_01DxMc1KFYqNjAwN3xojqRRx)_


deonmenezes:
Closing this as a duplicate: #161 (opened earlier the same day) already adds the same SQL-injection sink coverage to `source_sink_rules.js` for JS/Python, plus Go coverage this PR doesn't have, and uses a more robust quote-backreference regex for the concatenation rules (handles a SQL string with an embedded quote of the other kind, which this PR's simpler `[^'"]*` character class would miss).

`repo-checks / build-test`'s failure on this PR's head commit is also pre-existing and unrelated to this diff — it's the `stage_npm_packages.py` step trying to download native-build artifacts from a specific upstream `openai/codex` Actions run whose artifacts have since expired (`no valid artifacts found to download`), which #162 already root-caused as failing identically on `main` and on essentially every open PR right now. Not this PR's bug, and moot now that it's closing as a duplicate.

No further action needed here — see #161 for the surviving version of this change.

---

PR #160, opened 2026-09-13
fix: redact compound secret param names in http-audit evidence packs

deonmenezes:
## What gap this closes

`mantis_http_audit` (`.codex/mcp-servers/http-audit/server.js`) exists specifically to turn a captured HTTP exchange into a redacted evidence pack so raw secrets never land in a finding, artifact, report, or prompt (PRD section 9/11 — see `.codex/skills/http-evidence/SKILL.md`).

Its generic "secret-bearing param" redaction pattern was:

```js
/\b(token|api[_-]?key|access[_-]?token|refresh[_-]?token|secret|password|passwd|auth|session[_-]?id|sig|signature)=([^&\s]+)/gi
```

`\b` does not fire at an internal `_`/`-` boundary (both are word-adjacent), so this only matched *bare* param names like `token=` or `secret=`. Common **compound** param names used in real-world OAuth/session flows — `client_secret=`, `id_token=`, `user_password=`, `auth_token=`, `bearer_token=`, `api_token=` — silently passed through *unredacted* into the evidence pack's body preview.

Concretely, capturing a realistic OAuth token-exchange body before this fix:

```
grant_type=authorization_code&client_id=myapp&client_secret=SUPERSECRET123&redirect_uri=...
```

produced a preview with `client_secret=SUPERSECRET123` in the clear — exactly the raw-secret leakage this tool's whole purpose is to prevent.

## Why / fix

Added an optional `(?:[A-Za-z0-9]+[_-])?` prefix to the keyword alternation so compound names are matched (and the *whole* matched param name, prefix included, is preserved in the redaction output for readability). No other behavior changes.

## Verification

Ran the redaction regex against 21 cases:
- 16 "should redact" cases, including all the previously-missed compound names (`client_secret`, `user_password`, `id_token`, `bearer_token`, `api_token`) plus the pre-existing bare names (`token`, `secret`, `auth`, `sig`, `signature`, `session_id`, `passwd`, `password`, `api_key`, `access_token`, `refresh_token`) — all now redact correctly.
- 5 "should NOT redact" cases (`username=`, `sort_key=`, `authorization=Bearer...` as a body param, `page=2`, `category=auth-guide`) — confirmed no new false positives.

Also ran the fixed regex through the full `auditExchange()` handler against a realistic captured OAuth token request (headers + form body) end-to-end and confirmed the body preview now redacts `client_secret` correctly, and re-ran `prettier --check` on the file (already compliant, no reformatting needed).

[...]

PR #164, opened 2026-09-14
fix: close secret-detection gap between http-audit and findings-spine

deonmenezes:
## What gap this closes

Two places in the Mantis harness independently maintained "never let a raw secret escape" pattern lists, and they had drifted apart:

- `.codex/mcp-servers/http-audit/server.js` (`SECRET_VALUE_PATTERNS`) redacts AWS/GitHub/Slack key shapes, PEM private keys, JWTs, `user:pass@` URLs, **and** generic `token=`/`password=`/`secret=`/`api_key=`-style query/form params.
- `.codex/mcp-servers/findings/server.js` (`SECRET_PATTERNS`, the `scanForSecrets` DLP backstop that gates `finding_create`/`finding_update`) only checked the shape-based patterns — AWS/GitHub/Slack keys, PEM keys, JWTs — and was **missing** the generic named-secret pattern and the `user:pass@` URL pattern.

Net effect: a raw password or token pasted into a finding's `claim`, `location`, or `evidence` text (e.g. `"leak at /login?password=hunter2secret"`) would sail straight through `finding_create` even though the exact same text would have been redacted had it gone through `http_audit` first. That's a real false-negative in the one guard whose whole job (per the PRD invariant "never raw secrets/tokens/cookies/full bodies in findings/evidence") is to stop that.

## What changed

- Extracted the pattern list into a new shared module, `.codex/mcp-servers/lib/secret_patterns.js`, as the single source of truth (`SECRET_SHAPES`), with two accessors:
  - `detectionPattern(shape)` — non-global `RegExp` for `.test()` (safe to reuse across many calls, no `lastIndex` statefulness bugs).
  - `redactionPattern(shape)` — global `RegExp` for `.replace()`.
- `http-audit/server.js` now builds `SECRET_VALUE_PATTERNS` from the shared list instead of its own inline copy.
- `findings/server.js` now builds `SECRET_PATTERNS` from the same shared list, so it picks up the two patterns it was missing (generic named-secret params, `user:pass@` URLs) and can't drift out of sync with http-audit again.
- Added `.codex/mcp-servers/lib/secret_patterns.test.js` (Node's built-in `node:test` runner, zero new deps) covering: every shape is detected, the non-global detection regex is safely reusable across repeated calls, and redaction correctly strips the secret while preserving surrounding text/param names.

## Verification

- `node --test .codex/mcp-servers/lib/secret_patterns.test.js` — 4/4 pass.
- `npx prettier --check` on all touched/added files — clean.
- `node -c` syntax check on both modified server files — clean.
- Manual MCP stdio round-trip against the live `http-audit` server: request with `?password=...` and an `Authorization` header still comes back fully redacted — no behavior regression from the refactor.
- Manual MCP stdio round-trip against the live `findings` server:
  - `finding_create` with a claim containing `password=hunter2secret` → now correctly **refused** ("looks like it contains a raw secret/token/key") — previously this would have been silently accepted.
  - `finding_create` with a benign IDOR claim (no secret-shaped text) → still accepted normally, confirming no new false positives.

## Scope

Small and focused: no new dependencies, no changes to the lifecycle/transition logic, no changes to any agent prompts or skills — just closing the coverage gap in the shared secret-detection backstop and adding regression tests for it.

https://claude.ai/code/session_01RU2bAx1B5kxmGNBwetTuH7

---
_Generated by [Claude Code](https://claude.ai/code/session_01RU2bAx1B5kxmGNBwetTuH7)_


deonmenezes:
CI is red on this PR, but neither failing check is caused by this diff:

- **`repo-checks / build-test`** fails in `scripts/stage_npm_packages.py`, which tries to download native build artifacts from a hardcoded upstream `openai/codex` workflow run (`WORKFLOW_URL=".../openai/codex/actions/runs/26201494185"`) — `gh run download` returns `no valid artifacts found to download` because that run's artifacts have expired past GitHub's retention window. This job never touches `.codex/mcp-servers/**`.
- **`cargo-deny / cargo-deny`** fails on the `advisories` check — a pre-existing RUSTSEC advisory somewhere in the Cargo dependency graph. This PR contains no Rust changes at all.

I confirmed these aren't specific to this PR by checking `blocking-ci` run history: the same two checks fail identically on the last several unrelated PRs (#159 OSV/CVSS, #160 http-audit redaction, #161/#152 SQLi sinks, #162 JWT sinks), each touching completely different files. Neither has a fix available yet in any branch I can find, so there's nothing to port in. I'm not re-running the job — both failure modes are deterministic (expired GitHub artifact retention, a standing advisory), not flaky, and re-running has already been tried without effect across that same run history.

Given the same two checks are apparently blocking most/all of the ~60 open PRs in this repo right now (several duplicating each other's detection coverage because nothing merges), this looks like it needs a maintainer fix at the infra level: repoint `WORKFLOW_URL`/`CODEX_VERSION` in `scripts/stage_npm_packages.py` at a live artifact source, and either address or explicitly allow the standing `cargo-deny` advisory. Out of scope for me to fix blind from a 2-file JS PR. Leaving this PR open and watched.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RU2bAx1B5kxmGNBwetTuH7)_

---

PR #165, opened 2026-09-14
detect: add path-traversal (CWE-22) sink coverage to source_sink_scan

deonmenezes:
## What gap this closes

`source_sink_scan` (`.codex/mcp-servers/program-analysis/source_sink_rules.js`) is the fast regex proxy the Recon/Detect stages use for high-recall attacker-controlled-source / dangerous-sink surfacing (PRD FR-3.1/3.2) across JS/TS, Python, Go, and Java. It already covered command injection (CWE-78), XSS (CWE-79), unsafe deserialization (CWE-502), and (for Java) SQL injection (CWE-89) — but had **zero** rules for path traversal / arbitrary file read (CWE-22) in any of the four languages.

A classic vulnerable pattern like:

```js
fs.readFile(req.query.file, (err, data) => res.send(data));
```

or Flask's `send_file(request.args.get('path'))` produced no sink candidate at all, so this whole vulnerability class was silently invisible to the Detect stage's recall net — even though the exact same file already has a `js.http.query` *source* rule for `req.query`, there was no matching *sink* for it to pair with on this class.

Note: this is distinct from the "zip slip" archive-extraction variant of CWE-22 (paths escaping during zip/tar entry extraction) that's covered elsewhere — this closes the more common case of a request-controlled path flowing directly into a file-read/serve API.

## The fix

One sink rule added per language, matched to the same specificity as existing rules (dangerous framework/file APIs, not a blanket match on every `open()`/`os.path.join()` call, to avoid drowning recall in noise on codebases where those are ubiquitous and not attacker-influenced):

- **js**: `fs.readFile`/`readFileSync`/`createReadStream`/`writeFile`/`writeFileSync`, `res.sendFile`
- **py**: Flask `send_file`/`send_from_directory`
- **go**: `http.ServeFile`, `os.Open`/`os.ReadFile`, `ioutil.ReadFile`
- **java**: `new File(...)`/`new FileInputStream(...)`/`new FileReader(...)`

## Why it's safe / scoped

- Pure additive data in `RULES` — no change to the scan/matching engine (`server.js`) or existing rule behavior.
- Each new pattern is framework/API-name-anchored, matching the precision level of neighboring rules (e.g. `js.dom.innerhtml`, `java.objectinputstream`) rather than something broad enough to flag ordinary, non-attacker-influenced code.
- `every rule has the required shape` test in the new suite also regression-covers all 19 pre-existing rules (lang/kind/id/pattern/cwe shape + global-flag + uniqueness), so this run doubles as a light sanity check on the whole table, not just the new entries.

## Test plan

- [x] Added `.codex/mcp-servers/program-analysis/test_source_sink_rules.js` (Node's built-in `node:test`, zero deps, matching the repo's zero-dependency MCP-server philosophy) — 6 tests, all passing:
  - each new per-language rule fires on representative vulnerable snippets and not on unrelated code
  - full-table shape + uniqueness check
- [x] `node --test .codex/mcp-servers/program-analysis/test_source_sink_rules.js` → 6/6 pass
- [x] End-to-end: ran the live MCP server (`node .codex/mcp-servers/program-analysis/server.js`) via a raw `tools/call` for `source_sink_scan` against a fixture file containing `fs.readFile(req.query.file, ...)` — confirmed it now returns a `js.path_traversal.file_read_write` / `CWE-22` sink candidate where it previously returned none.
- [x] `prettier --check` clean on both changed files.

---

## Authorized discovery-only scan — this run

Also ran, as part of this run's recurring routine, a discovery-only scan against the single explicitly authorized target `https://vulnerable-bounty-site.vercel.app`. See the session summary for results (no separate report file this run — nothing new to persist beyond the summary; full detail in the final chat message).

---

https://claude.ai/code/session_012PAqqRSvmVM5A2Ygqcboao

---
_Generated by [Claude Code](https://claude.ai/code/session_012PAqqRSvmVM5A2Ygqcboao)_


deonmenezes:
Closing: this duplicates #108 and #132, both still open, both proposing the identical fix (CWE-22 path-traversal sink coverage for `source_sink_scan` across JS/Python/Go/Java). I opened this before checking the full open-PR list closely enough — my initial pagination missed #108/#132 — and only caught the overlap after pushing, while separately updating `reports/scans/vulnerable-bounty-site.md` and re-reading that file's own duplicate-tracking notes from prior runs.

No action needed here; #108 or #132 (whichever reviews cleaner) should be merged instead. Recommend a human triage pass across the growing pile of open, unmerged, and increasingly duplicate-prone detection PRs before this routine keeps proposing more.

---
_Generated by [Claude Code](https://claude.ai/code/session_012PAqqRSvmVM5A2Ygqcboao)_

---

PR #166, opened 2026-09-14
detect: add SQL injection sinks to source/sink heuristic scan (JS + Python)

deonmenezes:
## What

`Recon`/`Detect`'s `source_sink_scan` heuristic (`.codex/mcp-servers/program-analysis/source_sink_rules.js`) tags candidate dangerous sinks per language. Before this change, CWE-89 (SQL injection) sinks were only defined for **Java** (`statement.execute*`). JavaScript/TypeScript and Python — the two most common web-backend languages in the corpus this tool is likely to scan (Express/Flask/Django apps) — had **no SQL injection sink rules at all**, so dynamically-built SQL was invisible to the pipeline no matter how it was reached.

## Why this gap matters

SQL injection is one of the highest-impact, most common web vulnerability classes (OWASP Top 10, CWE-89). Missing it entirely for JS and Python means the Detect stage could walk right past a classic `db.query(\`SELECT * FROM users WHERE id = ${req.params.id}\`)` or `cursor.execute(f"SELECT * FROM users WHERE id = {user_id}")` without ever registering a candidate finding — a straight false negative in a stage whose whole job is "generous, high-recall" detection.

## Change

Adds two new sink rules, consistent with the existing heuristic style (regex proxy, not a taint engine — flags candidate call sites for the Validate stage to actually confirm reachability):

- `js.sql.dynamic_query` (CWE-89): flags `.query(...)` / `.execute(...)` calls built from template-literal interpolation (`` `...${x}` ``) or string concatenation (`"..." + x`).
- `py.sql.dynamic_query` (CWE-89): flags `.execute(...)` calls built from f-strings, `%`-formatting, or string concatenation.

Both rules are deliberately shaped to *not* flag the safe, parameterized form (`db.query("... = ?", [x])` / `cursor.execute("... = %s", (x,))`), keeping recall up without blowing up the false-positive rate the Validate stage would otherwise have to chew through.

## Testing

Added `.codex/mcp-servers/program-analysis/source_sink_rules.test.js` (Node's built-in `node:test`) covering both new rules: positive cases (template literal / concatenation / f-string / percent-formatting) and a negative case each (parameterized queries stay unflagged).

```
$ node --test .codex/mcp-servers/program-analysis/source_sink_rules.test.js
# tests 6
# pass 6
# fail 0
```

Also ran `npx prettier --check` on both changed files (repo already runs prettier over this directory per a prior commit) — no formatting changes needed.

## Scope

Small, additive, single-file rule change plus its test — no changes to the scan engine, the MCP tool surface, or any other detection class.

---

https://claude.ai/code/session_01WitTsoVhJr7RiKPny2tJrh

---
_Generated by [Claude Code](https://claude.ai/code/session_01WitTsoVhJr7RiKPny2tJrh)_


deonmenezes:
Closing as a duplicate, discovered only after pushing — my mistake for not checking the open PR list first.

This exact change (adding CWE-89 SQL-injection sink rules to `source_sink_rules.js` for JS/TS and Python) is already proposed, unmerged, in the open backlog at least 9 times: #101, #104, #114, #127, #139, #142, #152, #161, and now #166. None of these have merged — every one of the last several dozen PRs against this repo (going back to at least #107) fails the same `repo-checks / build-test` CI job for a reason unrelated to any of their diffs: `scripts/stage_npm_packages.py` tries to download native release artifacts from a hardcoded, now-expired upstream workflow run (`openai/codex` run `26201494185`, pinned `CODEX_VERSION=0.133.0-alpha.4`) and fails with "no valid artifacts found to download". That's a repo-wide CI/release-packaging issue, not something in scope for a source/sink-rule PR to fix.

Recommend the repo owner either (a) pick one SQLi PR from the list above to fix that pinned artifact reference and merge, closing the rest as duplicates, or (b) fix `stage_npm_packages.py`'s pinned workflow reference once at the base-branch level so the existing backlog can go green and get triaged.


---

PR #167, opened 2026-09-14
Add SQL-injection sink patterns to source_sink_scan (JS/Python/Go)

deonmenezes:
## What gap this closes

`mantis_program_analysis`'s `source_sink_scan` (the Detect-stage recall heuristic, `.codex/mcp-servers/program-analysis/source_sink_rules.js`) had **zero SQL-injection sink coverage outside Java's `statement.execute`**. SQL injection is a top-3 OWASP vulnerability class, and the scanner was structurally blind to it for JS/TS, Python, and Go targets — the three most common languages this tool is likely to be pointed at.

## What changed

Added CWE-89 sink patterns for the idiomatic unsafe-query shapes per language:

- **JS/TS**: `js.sql.template_query` — template-literal query with `${...}` interpolation passed to `.query(`/`.execute(`; `js.sql.concat_query` — string-concatenated query passed to the same call sites.
- **Python**: `py.sql.execute_fstring` — an f-string passed directly to `.execute(`; `py.sql.execute_concat` — `%`/`+`-built string passed to `.execute(`.
- **Go**: `go.sql.query_sprintf` — `fmt.Sprintf`-built query passed to `Query`/`QueryRow`/`Exec`(`Context`).

Patterns are deliberately scoped to the *unsafe construction* shape (interpolation/concatenation/formatting feeding the call), not just "any call to `.query`/`.execute`", so parameterized calls like `db.query("... = ?", [id])` do **not** match — keeping this consistent with the existing high-recall-but-not-noisy style of the other rules (e.g. the `yaml.load_unsafe` negative-lookahead for `SafeLoader`).

Also updated `.codex/skills/program-analysis/SKILL.md`'s tool description to mention the new coverage.

## Why

This is a pure-regex, dependency-free recall proxy per the PRD's "detection is generous" stage — it doesn't replace the reachability/validate stages, it just needs to *surface* candidates. Silently missing an entire top-tier vulnerability class for 3 of 4 supported languages was a coverage gap worth closing incrementally, without touching the SMT/AST tooling or the findings lifecycle.

## Testing

No existing test harness covers `source_sink_rules.js`. Verified manually with a smoke script running the new rules' regexes against both vulnerable and safe (parameterized) snippets for all three languages — all 9 cases passed, including 3 negative cases confirming no false positives on parameterized queries. Also ran `prettier --check` on the modified files (clean).

---

## Part 2 of this run: authorized discovery-only scan (blocked)

This run's second task called for a discovery-only scan against `https://vulnerable-bounty-site.vercel.app`. This session's outbound network egress policy denied the connection outright (proxy returned a 403 on CONNECT — a policy decision, not a transient error), so no scan traffic reached the target this cycle. No code change resulted from this half of the task; see my final message in this session for the network-policy note.

https://claude.ai/code/session_01NAdgmtZUSQQhX69N7NZuPT


deonmenezes:
Both `blocking-ci` failures on this PR's head (`74432e8`) are pre-existing infrastructure issues, not caused by this diff (which only touches `.codex/mcp-servers/program-analysis/source_sink_rules.js` and `.codex/skills/program-analysis/SKILL.md`):

**`cargo-deny / cargo-deny`** — fails in the `advisories` check. This reproduces identically on `main`'s own last `blocking-ci` run ([run 28905675542](https://github.com/deonmenezes/mantishack/actions/runs/28905675542/job/85752119003), the merge of #106): `advisories FAILED, bans ok, licenses ok, sources ok`. This PR contains zero Rust/Cargo changes, so it can't be the cause — the base branch is already red on this check.

**`repo-checks / build-test`** — fails at the "Stage npm package" step. Root cause: `scripts/stage_npm_packages.py` tries to `gh run download` native build artifacts from a hardcoded upstream reference pinned in `.github/workflows/repo-checks.yml` (`WORKFLOW_URL=https://github.com/openai/codex/actions/runs/26201494185`), and that run's artifacts have since expired — `no valid artifacts found to download`. This same job passed on `main`'s last run two months ago (before that expiry), and this PR touches nothing related to npm packaging or workflow pins. I don't have access to `openai/codex` from this session to identify a currently-valid replacement run, and re-pinning `CODEX_VERSION`/`WORKFLOW_URL` to a new upstream release is a maintainer call rather than something I should guess at and push blind — so I'm flagging it rather than widening this PR with a speculative fix. The underlying fix is bumping that pin to a `rust-release` run whose artifacts are still within GitHub's retention window.

No action needed on this PR's own diff for either failure.


---

PR #168, opened 2026-09-15
detect: add LDAP injection (CWE-90) sink coverage to source_sink_scan

deonmenezes:
## What gap this closes

`mantis_program_analysis`'s `source_sink_scan` heuristic (`.codex/mcp-servers/program-analysis/source_sink_rules.js`) had **zero LDAP injection (CWE-90) coverage** in any supported language. Nearly every other major injection class already has a rule or an open proposal (SQLi, NoSQL, SSTI, SSRF, XXE, CRLF, zip-slip, prototype pollution, open redirect, JWT bypass — see the repo's existing open PR queue), but LDAP filter injection — a classic top-10 web vuln for anything backed by an LDAP/AD directory (login forms, user search, group lookups) — was never covered.

I checked the open-PR queue first to avoid piling onto the existing duplication (multiple open PRs already re-propose the *same* SQLi/secret-redaction gaps); LDAP injection is not proposed anywhere in it, so this is net-new coverage rather than a duplicate.

## What changed

Added two CWE-90 sink rules, scoped to avoid the obvious false-positive traps for `.search(`-named methods:

- **`py.ldap.search`**: python-ldap's `search_s`/`search_ext_s`/`search_ext` — these suffixes are distinct enough from `re.search()`/`str.search()` that there's no realistic collision.
- **`js.ldap.search_filter`**: ldapjs's `client.search(dn, { filter: ... })` shape, gated specifically on the options-object `filter:` key so it does **not** fire on `String.prototype.search(regex)` or unrelated `.search()` client APIs (e.g. search-index/Elasticsearch clients) that take different argument shapes.

## Testing

No existing test harness covers `source_sink_rules.js` (consistent with prior PRs against this file). Verified manually:
- `node -e` sanity check of both regexes against synthetic true-positive (concatenated/template-built LDAP filter passed to the sink) and false-positive (`re.search`, `"str".search(/re/)`, an Elasticsearch-style `.search({...})` call) samples — all behaved as expected.
- End-to-end smoke test running the actual rule objects against synthetic `.py`/`.js` files containing both a vulnerable LDAP query and an unrelated `re.search`/`.search(/regex/)` call in the same file — only the vulnerable lines matched.
- `prettier --check` on the modified file — clean.

## Part 2 of this run: authorized discovery-only scan (blocked again)

Outbound access to `https://vulnerable-bounty-site.vercel.app` is still denied at the network-policy layer in this session (`curl` → `CONNECT tunnel failed, response 403`; agent-proxy status confirms `connect_rejected` / "gateway answered 403 to CONNECT (policy denial)" for that exact host). This is the same block prior runs hit (see the "blocked scan attempt" PRs from this repo's history) — it has not been resolved in the ~2 months this routine has been running. No scan traffic reached the target this cycle; this is an environment/organization network-policy setting (not something fixable from a code change in this repo) — it needs the scheduled environment's egress policy updated to allow that host. Flagging this directly to the repo owner outside of GitHub as well, since it means Part 2 of this recurring task has never actually been able to run.

---

https://claude.ai/code/session_011un9yYRCrpHctZJpXYZHdb

---
_Generated by [Claude Code](https://claude.ai/code/session_011un9yYRCrpHctZJpXYZHdb)_


deonmenezes:
CI status on this PR's head (`05daa6d`) — none of the current red/failing checks are caused by this diff, which only touches `.codex/mcp-servers/program-analysis/source_sink_rules.js`:

**`repo-checks / build-test`** — fails at the same "Stage npm package" step diagnosed on #167 today: `scripts/stage_npm_packages.py` runs `gh run download --repo openai/codex 26201494185`, and that upstream run's artifacts have expired (`no valid artifacts found to download`). This PR touches nothing related to npm packaging or workflow pins, so it can't be the cause. Re-pinning `WORKFLOW_URL`/`CODEX_VERSION` in `.github/workflows/repo-checks.yml` to a currently-valid `openai/codex` release is a maintainer call, not something to guess at from this session.

**`cargo-deny / cargo-deny`** — was still running as of this comment; per #167's identical diagnosis a few hours ago it fails on the `advisories` check and reproduces the same way on `main`'s own last run, unrelated to any Rust/Cargo changes (this PR has none).

**`Vercel` / `Vercel Preview Comments`** — this is the `opentradex-dashboard` project's own deployment check, an unrelated project attached to this GitHub account/org that runs on every PR in this repo regardless of diff content; not something this PR can affect.

The gate that actually reflects this diff, `rust-ci / CI results (required)`, is green. No action needed on this PR's own change for any of the above.

---

PR #169, opened 2026-09-15
detect: add SQL-injection sink patterns to source/sink heuristic scan

deonmenezes:
## Summary

`source_sink_scan` (the `mantis_program_analysis` MCP server's fast heuristic recall pass, used in the Recon/Detect stages of the Mantis pipeline) already tags dangerous sinks for `eval`/`exec`, command injection, XSS, and unsafe deserialization across JS and Python — but had **zero SQL-injection (CWE-89) coverage** in either language. Since SQLi is one of the most common vuln classes the Detect stage is supposed to surface (and Java already had a `java.statement.execute` sink), this was a real coverage gap in the "detect-generous" recall pass, not just a stylistic gap.

## What changed

Added two new sink rules to `.codex/mcp-servers/program-analysis/source_sink_rules.js`:

- `js.sql.string_built_query` (JS/TS): flags `.query(...)` / `.execute(...)` calls built from string concatenation or template-literal interpolation (`` `...${x}...` ``).
- `py.sql.string_built_query` (Python): flags `.execute(...)` / `.raw(...)` calls built from f-strings, `%`-formatting, or string concatenation.

Both are tagged `cwe: "CWE-89"` and registered as `sink` candidates, consistent with the existing rule shape/lifecycle (they still just feed the Detect stage's `candidate` list — reachability/validation stays a separate stage, per the `mantis-pipeline` skill).

Deliberately **not** flagged: calls using placeholder-style parameterized queries (e.g. `db.query("... WHERE id = ?", [userId])` or `cursor.execute("... WHERE id = %s", (user_id,))`), since the query text itself isn't attacker-built in that pattern — keeping the tool's existing high-recall/low-noise balance rather than flagging every `.query()`/`.execute()` call site indiscriminately.

Also updated `.codex/skills/program-analysis/SKILL.md`'s one-line description of `source_sink_scan`'s sink coverage to mention the new SQL sinks.

## Testing

- Verified both new regexes against hand-written vulnerable/safe JS and Python fixtures (string-concat/f-string/`%`-format/template-literal builds all match; parameterized-placeholder calls and unrelated `.query()`/`.execute()` calls don't).
- Ran the actual `mantis_program_analysis` MCP server over stdio (`tools/call` → `source_sink_scan`) against those fixtures end-to-end and confirmed it returns the two SQLi sinks and no false positives on the safe variants.
- `npx prettier --check` on both changed files — clean, no formatting drift.

[...]

PR #170, opened 2026-09-15
fix: surface subprocess timeouts across scanner wrappers instead of masking them

deonmenezes:
## What gap this closes

`.codex/mcp-servers/lib/run_tool.js`'s `runCommand` already tracks `timedOut` when a scanner subprocess (`semgrep`, `bandit`, `trivy`, `osv-scanner`, `codeql`, `ast-grep`) hits its `timeoutMs` budget and gets `SIGKILL`ed — but **no caller ever reads that flag**. Two silent failure modes result, both of which matter specifically for a vulnerability-detection tool that promises "Detect is generous... never fabricates":

1. **Timeout masquerading as a generic crash.** When a killed process's truncated stdout fails `JSON.parse`, every wrapper (`semgrep_scan`, `bandit_scan`, `trivy_scan`, `osv_scan`, `codeql_analyze`, `codeql_create_database`) returned the same ambiguous `"<tool> exited <code> and did not return parseable JSON"` whether the binary crashed, was misconfigured, or simply ran out of time on a big target. An operator can't tell "fix your config" from "raise the timeout and re-run" without digging into stderr/exit codes by hand.
2. **Timeout masquerading as a clean scan.** `ast_grep_scan` has no parse-failure branch at all — on a timeout it silently returns whatever partial `matches` had been parsed from the truncated stream as `{ available: true, match_count: N }`, indistinguishable from a real, complete result. A large repo that blows the 120s budget currently reports as a normal (possibly `match_count: 0`) scan instead of an incomplete one — a real false-negative risk baked into the harness itself, not the target.

## What changed

- `run_tool.js`: added a small shared `timeoutNote(result, toolName, timeoutMs)` helper (exported alongside `runCommand`/`notFoundMessage`) that returns `null` when a run didn't time out, or a clear, consistent "this is an incomplete/truncated scan, not a clean or empty result" message when it did — so every wrapper says the same thing instead of each re-deriving its own wording.
- Wired it into every consumer of `runCommand` under `.codex/mcp-servers/`:
  - `semgrep`, `bandit`, `trivy`, `osv-scanner`: the JSON-parse-failure branch now prefers the timeout message over the generic "did not return parseable JSON" one, and adds a `timed_out: true` field; the *successful*-parse path also carries a `warning` field for the (rarer but real) case where results were produced but the process was still killed by the timeout, so results shouldn't be assumed complete.
  - `codeql_create_database` / `codeql_analyze`: same pattern on their non-zero-exit branches.
  - `program-analysis`'s `ast_grep_scan`: added the same `warning` field on its previously-unconditional success return, so a truncated ast-grep stream is no longer indistinguishable from a genuine "no matches."
  - `program-analysis`'s `smt_check_reachability` was **not** touched — it already folds a timeout into its generic `"unknown"` verdict path with explicit `"z3 could not decide (unknown/timeout)"` wording, so there was no real gap there.

This repo currently has ~35 open PRs from prior runs of this same routine (mostly `source_sink_scan` sink-coverage additions and `http-audit`/findings-spine secret-redaction fixes). I checked all of their titles and diffs before starting, and none touch `run_tool.js`, timeout handling, or the scanner wrappers' error-reporting paths — confirmed with `search_pull_requests` for `timeout`/`timedOut`/`run_tool`, all zero hits. This is a fresh, narrow, cross-cutting reliability fix rather than another sink-coverage or redaction duplicate.

**Separately, for whoever reviews this:** this recurring routine has accumulated a large backlog of open, unmerged PRs (many are near-duplicates of each other proposing the same SQL-injection sink coverage, for example) going back to mid-July. Since nothing is being merged, each new run has to search a growing pile to avoid re-proposing the same fix, and the pile itself is starting to look like it needs a human pass (merge the good ones, close the superseded duplicates) rather than more automated additions on top. Flagging this here since it's outside the scope of a single incremental PR to fix.

## Testing

```
$ node --test .codex/mcp-servers/lib/run_tool.test.js
# tests 5
# pass 5
# fail 0
```

New `run_tool.test.js` (Node's built-in `node:test`, no new dependency) covers: a fast command resolves normally with `timedOut: false`; a deliberately slow command (`setTimeout` past its `timeoutMs`) is killed and reports `timedOut: true`; a missing binary reports `notFound: true` and not a timeout; and `timeoutNote()`'s null and non-null cases.

Also:
```
$ node --check <each modified file>   # all OK
$ npx prettier --check .codex/mcp-servers/{bandit,codeql,osv-scanner,program-analysis,semgrep,trivy}/server.js .codex/mcp-servers/lib/run_tool.js .codex/mcp-servers/lib/run_tool.test.js
```

None of the sibling scanner wrappers have a running external binary in this sandbox (they're all "report-until-installed" per `MANTIS.md`), so I couldn't exercise a live timeout end-to-end through e.g. `semgrep_scan` itself — the new test instead validates the underlying `runCommand`/`timeoutNote` mechanics directly, which is what every wrapper now depends on, following this layer's existing manual/unit-test convention (`findings/server.js` is the only other file here with test coverage, from still-open PR #156).

## Scope

Small and focused: one new shared helper, six wrapper files updated to use it, one new test file. No schema changes, no new dependencies, no behavior change on the non-timeout path.

---

## Part 2 of this run: authorized scan against `vulnerable-bounty-site.vercel.app`

Per the recurring routine, I attempted the authorized discovery-only scan against `https://vulnerable-bounty-site.vercel.app`. Same result as every prior run (see PRs #126, #131, #136, and #156's body): this environment's outbound network policy rejects the connection at the proxy/gateway level before any TLS handshake reaches the origin:

```
$ curl -sS -m 15 https://vulnerable-bounty-site.vercel.app
curl: (56) CONNECT tunnel failed, response 403
```

`$HTTPS_PROXY/__agentproxy/status` confirms a fresh `connect_rejected` entry for `vulnerable-bounty-site.vercel.app:443` timestamped this run (2026-09-15). No request reached the target, so there are no scan findings to report. This has now failed identically across multiple runs spanning July–September; it looks like the environment's network egress policy needs to explicitly allow this host before this half of the routine can ever produce a real result. I did not open another duplicate scan-log PR for this (several already exist, open and unmerged, saying the same thing).

https://claude.ai/code/session_01MRaoz1VJKHb8Fxg2jVy2zf

---
_Generated by [Claude Code](https://claude.ai/code/session_01MRaoz1VJKHb8Fxg2jVy2zf)_


deonmenezes:
`repo-checks / build-test` is failing on this PR, but it's not caused by this diff (which only touches `.codex/mcp-servers/*.js`).

**Root cause:** `.github/workflows/repo-checks.yml` lines 57-60 hardcode a specific upstream `openai/codex` workflow run to source pre-built native binaries from:
```yaml
CODEX_VERSION=0.133.0-alpha.4
WORKFLOW_URL="https://github.com/openai/codex/actions/runs/26201494185"
```
That run's artifacts have expired/rotated (`stage_npm_packages.py` reports `no valid artifacts found to download`, and `gh run download` then exits non-zero). This is a pre-existing, base-branch-level break, not something introduced here: the identical `repo-checks / build-test` failure is present on PR #169 (untouched by this change, same `main` base), confirming it fails independent of this diff.

I don't have access to the `openai/codex` repository from this session (my GitHub access here is scoped to `deonmenezes/mantishack` only), so I can't look up a current valid run to source a replacement `WORKFLOW_URL`/`CODEX_VERSION` pin, and fixing this is unrelated to the detection-pipeline change in this PR, so I'm not widening this PR's scope to patch it. Whoever has `openai/codex` access will need to update those two lines in `repo-checks.yml` to a workflow run that still has its native-artifact uploads live.

The other failing status on this PR (`Vercel` / `opentradex-dashboard`) is likewise unrelated to this repo's code — it's the same account-wide Vercel project integration failing identically on PR #169 and other open PRs, not something this change touches.

The rest of this PR's own checks (`rust-ci / CI results (required)`, `Codespell`, `Blob size policy`, `Detect changed areas`) are green.

---

deonmenezes:
One more pre-existing failure on this PR's head, same story as above: `cargo-deny / cargo-deny` fails with `advisories FAILED, bans ok, licenses ok, sources ok`. This PR contains zero Rust/Cargo changes (no `Cargo.toml`/`Cargo.lock`/`*.rs` touched), and the identical `cargo-deny` job fails with the exact same `advisories FAILED` signature and dependency tree on PR #169 (also untouched by any Rust change, same `main` base) — so this is a base-branch RUSTSEC-advisory issue in the existing `Cargo.lock`, not something this diff introduced or can fix by itself. Since this is Rust dependency-advisory triage (updating/pinning/allow-listing an advisory in `deny.toml` or `Cargo.lock`) unrelated to the JS detection-tooling change here, I'm leaving it to whoever owns the Rust dependency graph rather than widening this PR's scope.

With both `repo-checks / build-test` and `cargo-deny` confirmed pre-existing/base-branch failures (not caused by this diff) and this PR's own relevant checks green, I'm not spending the one allowed re-run on either — they're not flakes, they reproduce identically on an unrelated PR's separate CI run.

---

PR #174, opened 2026-09-16
fix: redact JSON-quoted secrets in http-audit evidence packs

deonmenezes:
## What gap this closes

`mantis_http_audit` (`.codex/mcp-servers/http-audit/server.js`) is the pure-Node evidence-pack tool used during Validate/DAST to attach reproducible HTTP request/response evidence to a finding without leaking raw secrets (PRD section 9/11 invariant: "never raw secrets/tokens/cookies/full bodies" in findings evidence).

Its generic secret-value redaction only matched **query-string shape** (`token=abc123`). Any secret in a **JSON body** — `"api_key": "abc123"`, `"access_token": "..."` — sailed straight through unredacted. That's a false negative on the most common real-world shape: modern web/API targets return JSON, not form-encoded bodies, so the exact case the redactor exists to prevent (a finding's attached evidence pack containing a live token straight from a captured response) was not covered.

This also left several common credential *shapes* undetected: Google API keys (`AIza...`), Stripe keys (`sk_live_...`/`sk_test_...`), npm tokens (`npm_...`), and a bare `Bearer <token>` appearing outside the `Authorization` header (e.g. copy-pasted into a body/log line).

## What changed

- Added a JSON-quoted variant of the generic secret-param pattern (`"token": "..."`, `"api_key": "..."`, etc.) alongside the existing query-string one.
- Added shape-based patterns for Google API keys, Stripe keys, npm tokens, and inline `Bearer <token>`.
- Guarded the `createServer(...)` stdio bootstrap behind `require.main === module` so the pure redaction functions (`redactValue`, `parseHttpMessage`, `auditExchange`) are importable and unit-testable — this file previously had zero test coverage. No behavior change to the MCP server itself (verified by hand over stdio, see test plan).
- Added `.codex/mcp-servers/http-audit/server.test.js` with `node:test` coverage for the new patterns and for the JSON-body redaction path end-to-end through `auditExchange`.

Scope is intentionally narrow: no changes to header redaction, request-ref hashing, or any other tool.

## Test plan

- [x] `node --test .codex/mcp-servers/http-audit/server.test.js` — 3/3 pass
- [x] `npx prettier --check .codex/mcp-servers/http-audit/*.js` — clean
- [x] Manual stdio smoke test: piped `initialize` + `tools/call http_audit` with a JSON body containing a `token` field through `node .codex/mcp-servers/http-audit/server.js` directly and confirmed the MCP server still responds correctly and the field comes back as `[REDACTED]` in the evidence pack.

---

Part of the recurring Mantis detection-maintenance routine (mantishack repo, run 2026-09-16). Also ran an authorized discovery-only scan against the repo owner's designated practice target this run; see session summary for findings (no code change resulted from the scan itself).

https://claude.ai/code/session_01D2SyypW3HFyudTf8S82xbh

---
_Generated by [Claude Code](https://claude.ai/code/session_01D2SyypW3HFyudTf8S82xbh)_


deonmenezes:
`repo-checks / build-test` is failing on this PR, but it's not caused by this diff (which only touches `.codex/mcp-servers/http-audit/*.js`).

**What's failing:** the job's npm-packaging step in `.github/workflows/repo-checks.yml` (added in d155094, 2026-07-06, untouched since) shells out to `scripts/stage_npm_packages.py` with a hardcoded `WORKFLOW_URL="https://github.com/openai/codex/actions/runs/26201494185"` to pull prebuilt native binaries from that fixed upstream `openai/codex` Actions run. That run's `x86_64-unknown-linux-musl` artifact download now returns "no valid artifacts found to download" (GitHub Actions artifacts expire; this pinned run is over two months old), so `gh run download` exits 1 and the job fails before any of this PR's own code is touched.

**Why I'm not fixing it here:** this is unrelated to the diff, and I don't have access to the `openai/codex` repo from this session to find a current run with live artifacts to re-pin to — picking a replacement run id blind would just trade one stale pin for another. This same failure is already root-caused identically on unrelated PRs going back to at least #161 (2026-09-13) and reconfirmed on #162, #170 — it reproduces on every PR regardless of diff, so it's very likely red on `main` too, not something this PR introduced.

**Proposed fix (for whoever picks this up):** either (a) re-pin `WORKFLOW_URL` to a current `openai/codex` release run with live artifacts, or (b) make the artifact-download step tolerant of a missing/expired upstream run (skip/soft-fail with a warning) so this job's fate isn't hostage to an external repo's artifact retention window.

No action needed on my end beyond this note — I don't have the means to re-run/confirm-fix a check outside this PR's own diff scope.

---

deonmenezes:
A second check also failed on this PR: `cargo-deny / cargo-deny`. Also not caused by this diff.

**What's failing:** the `advisories` check in `cargo-deny` flags `webbrowser 1.0.6` (pinned in `codex-rs/Cargo.lock`, pulled in directly by `codex-login`, `codex-rmcp-client`, and `codex-tui`) for **RUSTSEC-2026-0257 / GHSA-2ph8-5cr8-hr33**: on Unix, an attacker-controlled non-HTTP(S) URL with spaces can inject extra arguments into the `BROWSER` env-var launch command (`--remote-debugging-port`, `--proxy-server`, etc.). `bans`, `licenses`, and `sources` all pass — only `advisories` fails.

**Fix (verified available, not yet applied anywhere):** the advisory's own solution is `cargo update -p webbrowser` — fixed in 1.2.2+, and 1.2.4 is current on crates.io (checked directly). I didn't apply it in this PR: it's an unrelated Cargo workspace dependency bump across 3 crates in a ~100-crate Rust workspace, which needs a full `cargo`/`cargo-deny` run to verify before pushing (I've verified this specific PR's own diff with `node --test` + `prettier`, not the Rust toolchain), and bundling it here would widen this PR well past "JSON secret redaction in http-audit." No existing open PR does this update either (checked).

Same pattern as the `repo-checks / build-test` failure above: pre-existing, diff-unrelated, reproduces on this PR only because it reproduces on every PR right now. Flagging both here per this PR's CI, and separately to the repo owner as part of this run's summary — this one in particular (an unpatched RUSTSEC advisory sitting unfixed across dozens of PRs) seems worth a dedicated fix PR from a session that can run the full Rust toolchain to validate the bump.

---

PR #176, opened 2026-09-18
detect: add ReDoS (CWE-1333) sink coverage to source_sink_scan

deonmenezes:
## What this closes

Part of the recurring 4-hour maintenance routine's detection-improvement task. `source_sink_rules.js` had no coverage for dynamic regular-expression construction, a recognized precursor to both regex injection and attacker-tunable catastrophic backtracking (ReDoS, CWE-1333). Cross-checked the ~75 open PRs on this repo first and none of them touch regex/ReDoS, so this is a genuinely new gap rather than a duplicate of existing detection-coverage PRs.

## What this adds

Two new heuristic sink rules in `.codex/mcp-servers/program-analysis/source_sink_rules.js`:

- `js.regexp.dynamic_construction` — flags `new RegExp(...)` when the first argument is **not** a string/template literal.
- `py.regex.dynamic_construction` — flags `re.compile/match/fullmatch/search/sub/split(...)` when the first argument is **not** a string literal (accounting for Python's `r`/`b`/`u`/`f` string prefixes, e.g. `re.compile(r"^[a-z]+$")` correctly doesn't fire).

Both stay off when the pattern is a static, human-authored literal, which is the overwhelming majority of real calls — the goal was closing the coverage gap without adding false-positive noise to the candidate list, consistent with the existing rules' "heuristic proxy, not a taint engine" design.

## Verification

- `node -c` on the edited file (syntax) and `npx prettier --check` (repo formatting) both pass — no existing test suite covers this heuristic module (none previously existed for `source_sink_rules.js`).
- Hand-verified via a small fixture: `new RegExp(userPattern)` / `re.compile(pattern)` correctly fire at the right line; `new RegExp("^[a-z]+$")` / `re.compile(r"^[a-z]+$")` / `re.search("foo", text)` correctly do not.

## Second task this run: authorized scan against the bounty target

[...]

PR #175, opened 2026-09-16
detect: add hardcoded-credential (CWE-798) sink coverage to source_sink_scan

deonmenezes:
## What gap this closes

Part of the recurring 4-hour maintenance/testing routine's first task: find and close one incremental detection gap.

`source_sink_scan` (the dependency-free heuristic in `mantis_program_analysis`) had **zero working coverage for hardcoded secrets in source code**:

- `mantis_trufflehog` is the repo's only dedicated secret scanner, but it reports `available: false` until the `trufflehog` binary is installed in the execution environment — it currently detects nothing.
- `mantis_http_audit` redacts secrets, but only in *captured HTTP evidence* you hand it — it never scans source files.

So a codebase with a literal AWS key or password string committed in a `.js`/`.py`/`.go`/`.java` file currently sails through Detect with no candidate raised at all, regardless of environment.

## What changed

Added CWE-798 sink rules to `.codex/mcp-servers/program-analysis/source_sink_rules.js`, two per already-supported language (js/ts, py, go, java):

1. **`*.hardcoded_credential`** — a generic key-name/assignment shape (`password`, `passwd`, `secret`, `api_key`, `access_key`, `auth_token`, `private_key`, `client_secret`, ...) assigned a quoted literal of at least 8 chars.
2. **`*.aws_access_key`** — the distinctive `AKIA[0-9A-Z]{16}` AWS access-key-id shape, which needs no key-name context to be high-confidence.

Two deliberate design choices, both verified with a standalone test script against 18 true/false-positive cases:

- The key-name pattern has **no leading `\b`**, so it also matches compound/camelCase identifiers like `dbPassword` or `userApiKey` — not just bare `password =`. It still can't fire on non-trailing substrings (`passwordConfirmation = ...` doesn't match) because the key name must sit directly against the `:`/`=` operator either way.
- A short placeholder exclusion list (`changeme`, `example`, `your-api-key-here`, `test`, ...) keeps common config-sample snippets from dominating matches with noise, without needing a full allow/deny secret-entropy engine.

Also added one clarifying sentence to `.codex/skills/program-analysis/SKILL.md`: unlike the rest of the rule set, a CWE-798 finding doesn't need a source→sink dataflow trace to validate — the literal's presence in shipped source *is* the exposure, so Validate should confirm the value is a real/non-placeholder/non-already-rotated secret instead of chasing a reachability path.

## Testing

[...]

PR #177, opened 2026-09-18
fix: bump crossbeam-epoch to 0.9.21 to clear RUSTSEC-2026-0204

deonmenezes:
## What this closes

This is the recurring 4-hour maintenance routine, but instead of another detection-coverage PR I chased down why the last ~70 automation PRs (since #106, merged 2026-07-07) have all shown red required checks regardless of what they change.

Root cause: `cargo-deny`'s `advisories` check fails on `crossbeam-epoch v0.9.18`, pinned in `codex-rs/Cargo.lock` and pulled in transitively via `crossbeam-deque` (used by `codex-file-search` and others) and `moka`. It's affected by [RUSTSEC-2026-0204](https://rustsec.org/advisories/RUSTSEC-2026-0204.html) (invalid pointer dereference in the `fmt::Pointer` impl for `Atomic`/`Shared` when the pointer is null). That failure is environmental (the RustSec advisory DB, not the diff), so it reproduces identically on `main`'s current HEAD and on every PR built against it — it's the reason nothing has merged in ~10 weeks despite dozens of otherwise-mergeable detection PRs sitting open.

## What this does

Bumps the `crossbeam-epoch` pin in `codex-rs/Cargo.lock` from `0.9.18` to `0.9.21` (latest on crates.io, not yanked). The advisory is fixed upstream as of `>= 0.9.20`, so this is a real fix rather than another `deny.toml` ignore entry (the repo already has a documented ignore-list pattern in `deny.toml`/`.cargo/audit.toml` for advisories with *no* upstream fix yet — this one has a fix, so it doesn't belong there).

Confirmed via the crates.io API that `crossbeam-epoch` 0.9.21 declares identical dependency requirements to 0.9.18 (`crossbeam-utils ^0.8.18`, dev-only `rand ^0.8`, optional `loom ^0.7.1`), so this is a version+checksum-only change with no dependency-graph impact:

```
diff --git a/codex-rs/Cargo.lock b/codex-rs/Cargo.lock
 name = "crossbeam-epoch"
-version = "0.9.18"
+version = "0.9.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5b82ac4a3c2ca9c3460964f020e1402edd5753411d7737aa39c3714ad1b5420e"
+checksum = "dc74980687109a3b14c72fd458107bf0baa1da1a1a805e178d15501ba9b86d9d"
```

## Validation

This sandbox's egress policy blocks `chromium.googlesource.com`, which an unrelated workspace crate (`codex-realtime-webrtc`'s `libwebrtc` git submodule) needs to resolve — so any `cargo update`/`cargo check`/`cargo metadata` invocation over this workspace fails locally here regardless of which package it targets, before it ever gets to `crossbeam-epoch`. That's a sandbox network-policy limitation, not something this change touches.

I verified the fix out-of-band instead:
- `crates.io` version list: `0.9.21` exists and is not yanked.
- `crates.io` per-version dependency list for `0.9.18` vs `0.9.21`: identical.
- `crates.io` checksum for `0.9.21`: matches what's now in the lockfile.
- The RustSec advisory text: confirms `patched = [">= 0.9.20"]`.

CI (with full network access) should be able to run the real `cargo update`/`cargo deny check advisories` and confirm the resolve is clean and the advisory clears. If it surfaces something this out-of-band check couldn't see, happy to adjust.

## Why this matters for the pipeline

Everything under "improve detection capabilities" is moot if nothing can merge. This doesn't add a detection rule, but it's the highest-leverage fix available this run: it's the actual blocker sitting in front of the ~70 other open automation PRs (several of which are genuine, non-duplicate detection-coverage improvements — e.g. #176 ReDoS, #175 hardcoded credentials, #158 SSRF, #151 prototype pollution). Worth noting separately: a chunk of that backlog (#152/#154/#161/#163/#166/#167/#169/#171) is duplicate SQL-injection sink-coverage work generated independently across different runs against the same lines of `source_sink_rules.js` — that's a separate cleanup the repo owner may want to triage (closing the redundant ones) once merges are flowing again.

---

https://claude.ai/code/session_01167iJTE2QEsS7mfXJVTFJQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01167iJTE2QEsS7mfXJVTFJQ)_


deonmenezes:
## CI status on `4a2ac4c`

**Confirmed: the fix in this PR works.** The `cargo-deny` advisories report no longer lists `RUSTSEC-2026-0204`/`crossbeam-epoch` at all — that part of the original failure is resolved.

**But `cargo-deny / cargo-deny` is still red**, for a reason unrelated to this diff: a second, independently-published advisory, `RUSTSEC-2026-0257` (`webbrowser` v1.0.6, an argument-injection issue via the `BROWSER` env-var template on Unix, published 2026-07-29, fixed in >=1.2.2), is now the sole `advisories FAILED` cause (`bans ok, licenses ok, sources ok`). This PR never touches `webbrowser`, and this will fail identically on `main` and on every other open PR, same as the crossbeam-epoch issue did — it's the live RustSec DB, not this diff. I checked: no other open PR already fixes it.

I'm not pushing a fix for it into this PR. Bumping `webbrowser` 1.0.6 -> 1.2.2+ isn't a same-shape edit to `crossbeam-epoch`'s: crates.io shows the newer `webbrowser` releases drop `core-foundation`, bump `jni` to `^0.22.2`, and add a brand-new normal dependency (`objc2-app-kit ^0.3`, macOS-only) that isn't in the lockfile at all today. That means new package entries (version, checksum, transitive deps) would have to be added, not just a version+checksum swap on an existing entry — not something I can hand-verify against crates.io metadata with the same confidence, and this sandbox can't run `cargo update` itself (see the PR description: it's blocked by an unrelated git-submodule host). Proposed patch for whoever/whatever run has real `cargo` network access: `cd codex-rs && cargo update -p webbrowser --precise 1.2.2` (or later 1.2.x), then re-run `cargo deny check advisories` to confirm clean.

Separately, `repo-checks / build-test` is also red on this PR's head, but again not because of this diff: the `Stage npm package` step in `.github/workflows/repo-checks.yml` hardcodes `WORKFLOW_URL="https://github.com/openai/codex/actions/runs/26201494185"` to fetch native-binary artifacts for an npm-packaging smoke test, and that specific upstream workflow run's artifacts are gone (`no valid artifacts found to download`). Same failure, same command, same exit code on #176 (independent diff, unrelated to Rust deps) — so this is a third, separate, pre-existing environmental blocker: a pinned external artifact reference that was always going to expire (GitHub's default retention is 90 days) with no refresh mechanism. Worth a follow-up PR pointing that step at a still-valid run (ideally this fork's own release artifacts instead of upstream's), but that's out of scope for this Cargo.lock-only PR.

Net: this PR is correct and necessary, but not sufficient on its own to get `main` fully green again — two more independent, pre-existing issues (the webbrowser advisory and the stale artifact reference) also need fixing. I'll keep watching this PR.

---

https://claude.ai/code/session_01167iJTE2QEsS7mfXJVTFJQ

---

PR #178, opened 2026-09-18
Add SQL injection sink detection to source_sink_scan (JS/Python/Go)

deonmenezes:
## What gap this closes

`mantis_program_analysis`'s `source_sink_scan` heuristic (`.codex/mcp-servers/program-analysis/source_sink_rules.js`) is the Recon/Detect-stage recall tool for attacker-controlled sources and dangerous sinks across JS/TS, Python, Go, and Java. Its CWE-89 (SQL injection) coverage was **Java-only** (`java.statement.execute`) — JS, Python, and Go had zero SQLi sink rules, even though SQL injection is explicitly called out as a target vulnerability class in the pipeline (`mantis-pipeline` SKILL.md) and is one of the most common, highest-impact bug classes scanners target.

That meant a Recon/Detect pass on any JS, Python, or Go target would surface command injection, XSS, deserialization, etc., but silently skip raising SQLi candidates at all for those three languages — a real coverage gap, not just weak signal.

## What changed

Added sink rules for the classic interpolated/concatenated SQL construction patterns per language:

- **JS/TS**: template-literal interpolation into `.query()/.execute()/.raw()` (`` db.query(`...${id}`) ``), string concatenation into `.query()/.execute()`, and raw-query ORM helpers (`knex.raw`, `sequelize.query`, `$queryRawUnsafe`).
- **Python**: f-string interpolation into `.execute()`, and string concatenation or `.format()` into `.execute()`.
- **Go**: `fmt.Sprintf`-built queries passed to `Query`/`QueryRow`/`Exec` (and their `Context` variants).

Rules key off the **dangerous construction shape** (interpolation/concatenation feeding the query string) rather than the bare API call, since `query()`/`execute()` are used interchangeably for both raw and parameterized queries in these ecosystems. This keeps precision high — parameterized calls like `db.query("... WHERE id = ?", [id])` are deliberately *not* flagged, verified by the new tests below.

Also updated `.codex/skills/program-analysis/SKILL.md`'s sink examples to mention SQL sinks.

## Testing

- Added `.codex/mcp-servers/program-analysis/source_sink_rules.test.js` (Node's built-in `node:test`, zero new deps) — this module had **no prior test coverage**. Covers true positives for every new rule and true negatives for parameterized/safe equivalents, plus two whole-ruleset sanity checks (unique rule ids, every sink rule carries a `cwe`).
- `node --test .codex/mcp-servers/program-analysis/source_sink_rules.test.js` → 8/8 passing.
[...]

PR #172, opened 2026-09-16
detect: add insecure-randomness (CWE-330) sink coverage to source_sink_scan

deonmenezes:
## What gap this closes

`source_sink_scan`'s heuristic rule table (`.codex/mcp-servers/program-analysis/source_sink_rules.js`) had zero coverage for **CWE-330 (Use of Insufficiently Random Values)**. `Math.random()` (JS), `random.random()`/`random.randint()`/`random.choice()`/`random.randrange()` (Python), and `new Random()` (Java) are not CSPRNGs, but code that uses them to mint password-reset tokens, session ids, OTPs, or CSRF/nonce values is a real and common vulnerability (predictable tokens are directly guessable/brute-forceable). Nothing in the rule set flagged this.

## What I checked first

Before touching anything I audited the ~67 currently-open PRs against this repo (many from prior runs of this same scheduled routine) to avoid adding yet another duplicate. I found **8+ open/closed PRs all proposing the same SQL-injection sink coverage** (#152, #154, #161, #163, #166, #167, #169, #171) and **~20 more** open PRs each claiming a different CWE (LDAP injection #168, SSTI #128, XXE #125, prototype pollution #115/#151, NoSQL injection #146, open redirect #147, SSRF #158, CRLF #157, Zip Slip #155, weak crypto #117, JWT bypass #162, etc.) that overlap heavily with each other. None of those are in this diff, and none of those CWEs are what this PR touches — I picked CWE-330 specifically because no open or closed PR covers it. (Flagging the duplicate-PR backlog separately to the repo owner — it's a real problem but out of scope for a single detection-improvement PR.)

## What changed

Added one sink rule per language (JS/TS, Python, Java), each scoped to **same-line co-occurrence** with a security-sensitive identifier (`token`, `session`, `password`, `otp`, `csrf`, `nonce`, `secret`, `api_key`, ...) rather than matching every `Math.random()` / `random.*()` / `new Random()` call. A bare match on those APIs would be extremely noisy (they're used constantly for jitter, dice rolls, shuffling, UI animation) and would drown the signal — consistent with the existing `py.yaml.load_unsafe` rule's precedent of using same-line context to cut false positives.

```
js.insecure_random_security_value    -> Math.random() near token/session/password/...
py.insecure_random_security_value    -> random.random|randint|choice|randrange() near token/session/...
java.insecure_random_security_value  -> new Random() near token/session/password/...
```

All three tag `cwe: "CWE-330"`.

## Verification

[...]

PR #179, opened 2026-09-18
fix: close secret-shape gap between findings-spine DLP guard and http-audit redactor

deonmenezes:
## What gap this closes

`.codex/mcp-servers/findings/server.js` (the findings-spine's `scanForSecrets` refusal guard on `finding_create`/`finding_update`) and `.codex/mcp-servers/http-audit/server.js` (the DAST evidence-pack redactor) each kept an independently-maintained copy of "what does a secret look like." They had drifted:

- http-audit's list additionally caught **`user:pass@` URL userinfo** and **generically-named secret query/form params** (`password=`, `api_key=`, `token=`, `secret=`, `session_id=`, etc.) via a shape-independent regex on the param name.
- the findings-spine guard only had the shape-based patterns (AWS keys, GitHub/Slack tokens, JWTs, PEM keys) — it was missing both of the above.

Net effect: an agent could call `finding_create`/`finding_update` with a `claim`/`evidence` payload embedding `admin:hunter2@host` or `password=hunter2`, and it would sail straight into the append-only findings log untouched — even though the exact same text would have been redacted had it instead passed through `http_audit` first. That's a real false-negative in the one enforcement point the PRD (section 9/11) says exists specifically to keep raw secrets out of findings.

## Fix

Extracted both pattern lists into a single shared module, `.codex/mcp-servers/lib/secret_patterns.js`, exporting:
- `redactValue(value)` — in-place redaction (used by http-audit), unchanged behavior.
- `containsSecret(value)` — boolean guard (used by findings), now covering the full pattern set. Rebuilds each regex without the `g` flag per call so repeated invocations can't trip on stale `lastIndex` state from a shared global-flagged `RegExp#test()` — a latent footgun the old `findings` code didn't have (its patterns weren't `g`-flagged) but the merged list would have hit.

Both servers now `require()` this shared list instead of keeping their own copy, so the two DLP enforcement points can't drift apart again.

## Verification

- `node -c` on all three touched files.
- `prettier --check` and `scripts/asciicheck.py` clean.
- Unit-level: `containsSecret`/`redactValue` exercised directly against AWS/GitHub/Slack/JWT/PEM/userinfo/generic-param secret shapes plus a clean string, including repeated calls to confirm no `lastIndex` state leak.
- End-to-end over the real MCP stdio protocol (spawned each server, drove `initialize` + `tools/call`):
  - `finding_create` with `claim: "curl https://admin:hunter2@example.com/api leaked in log"` → now **refused** (previously would have been accepted).
  - `finding_create` with `claim: "...password=hunter2 in cleartext query string"` → now **refused** (previously would have been accepted).
  - `finding_create` with a clean XSS claim (no secret shape) → still **succeeds** normally.
  - `http_audit` on a request with `token=` in the URL, an `Authorization` header, and a `Cookie` header → redaction output unchanged from before the refactor.

Scope kept intentionally small: no new secret shapes were invented, no lifecycle/schema changes — just parity between two guards that are supposed to enforce the same policy, plus the shared module to keep them that way.

---

https://claude.ai/code/session_01HUFLeHPQveKvT9kqEind8B

---
_Generated by [Claude Code](https://claude.ai/code/session_01HUFLeHPQveKvT9kqEind8B)_


deonmenezes:
Two required checks are failing on this PR's head (`268cb1a`), and both are pre-existing, repo-wide breakage on `main` unrelated to this PR's diff (3 JS files under `.codex/mcp-servers/`, no Rust/Cargo/workflow changes):

- **`cargo-deny / cargo-deny`** — fails on the `advisories` check: `crossbeam-epoch 0.9.18` (pinned in `codex-rs/Cargo.lock:305`, pulled in transitively via `crossbeam-deque` → `ignore` → `codex-file-search`) is flagged by [RUSTSEC-2026-0204](https://rustsec.org/advisories/RUSTSEC-2026-0204). This has been failing on every PR against `main` since #106 merged (2026-07-07), independent of what any individual PR changes — confirmed by checking `main`'s own history and several other recent, unrelated open PRs (#172, #174–176, #178), all showing the identical failure. A fix attempt exists at #177 (bumps the pin to `0.9.21`, where the advisory is patched upstream), but that PR's own CI run still shows `crossbeam-epoch 0.9.18` at the same `Cargo.lock:305` and still fails the same check — so its hand-edited lockfile change hasn't actually taken effect yet (its author noted this sandbox's egress policy blocked `cargo update`/`cargo check` over the full workspace, so it was never validated against a real resolve). There's nothing to port here yet; merging #177 as-is would not currently clear this.

- **`repo-checks / build-test`** — fails staging npm packages: `stage_npm_packages.py` tries to download native-binary artifacts from a hardcoded upstream `openai/codex` Actions run (`https://github.com/openai/codex/actions/runs/26201494185`), which returns "no valid artifacts found to download" (the run's artifacts have presumably since expired — GitHub Actions artifacts don't live forever). This is a hardcoded value in `.github/workflows/repo-checks.yml`/the staging script, not something this diff touches. I can't propose a working patch for it from this session: fixing it means pointing at a still-live `openai/codex` release workflow run, and `openai/codex` is outside this session's repository scope, so I have no way to look up a currently-valid run id.

Both are infra/dependency issues for the repo owner to unblock (landing #177 properly, and refreshing the pinned upstream artifact run), not something in scope for this PR to fix. I'm not re-running either check — they're fully deterministic (identical failure signature reproduced across dozens of runs over ~10 weeks, including on the PR specifically written to fix one of them), so a re-run wouldn't change the outcome.

---

PR #180, opened 2026-09-18
docs: record blocked discovery scan against the authorized bounty target

deonmenezes:
## Summary

The recurring maintenance routine's discovery-scan half (task 2: authorized, discovery-only scan of `https://vulnerable-bounty-site.vercel.app`) could not run this cycle — this session's outbound egress proxy denied the destination with a `403` (not on the environment's egress allowlist). Confirmed via two independent paths:

- `curl` through the local agent proxy: `CONNECT tunnel failed, response 403` / `connect_rejected -- organization policy`.
- The harness's own `WebFetch` tool: `EGRESS_BLOCKED` for the same domain.

The environment's own proxy documentation (`/root/.ccr/README.md`) is explicit that a `403`/`407` policy denial should be reported, not retried or routed around, so no scan traffic was sent and no other tool/workaround was attempted.

## What this adds

`reports/vulnerable-bounty-site.md` — a running, append-per-run scan log for this authorized target, so once egress is allowlisted, future routine runs have prior-run history to diff new/changed findings against (per the routine's own reporting requirement). This run's entry records the blocker and what the repo owner needs to do to unblock it (add the domain — and the other 9 authorized targets once provided — to this session's/environment's egress allowlist).

No scan findings to report this cycle: zero requests reached the target.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01HUFLeHPQveKvT9kqEind8B

---
_Generated by [Claude Code](https://claude.ai/code/session_01HUFLeHPQveKvT9kqEind8B)_

PR #181, opened 2026-09-19
fix: bound stdout/stderr buffering in scanner runCommand helper

### LINKED ISSUE #149: Bound scanner stdout and stderr buffering in runCommand
## Problem

`.codex/mcp-servers/lib/run_tool.js` appends every stdout and stderr chunk to strings until the child process exits. There is no byte limit on either stream.

All external scanner integrations use this helper, including Semgrep, CodeQL, OSV Scanner, TruffleHog, Bandit, Trivy, ast-grep, and Z3. A large repository, verbose scanner, or malfunctioning process can therefore make the MCP server retain an arbitrary amount of output and exhaust memory before individual handlers get a chance to parse or truncate it.

This also conflicts with the repository's bounded-context principle: several handlers truncate stderr only after `runCommand` has already buffered the complete stream.

## Reproduction

Run from the repository root with Node.js available:

```js
const { runCommand } = require("./.codex/mcp-servers/lib/run_tool.js");

(async () => {
  const bytes = 20 * 1024 * 1024;
  const result = await runCommand(process.execPath, [
    "-e",
    `process.stdout.write("A".repeat(${bytes})); process.stderr.write("B".repeat(${bytes}))`,
  ]);
  console.log({
    stdoutBytes: Buffer.byteLength(result.stdout),
    stderrBytes: Buffer.byteLength(result.stderr),
  });
})();
```

Observed result:

```text
{ stdoutBytes: 20971520, stderrBytes: 20971520 }
```

Increasing `bytes` increases retained memory without a bound.

## Proposed behavior

Add explicit per-stream byte limits to `runCommand` while continuing to drain both pipes so a child cannot block on a full pipe. Once a limit is reached, retain only the bounded prefix, record that truncation occurred, and return structured flags such as `stdoutTruncated` and `stderrTruncated`.

The default should protect every scanner. Callers that require a larger complete JSON or SARIF document could request a documented larger bound, or write scanner output to a temporary file and parse it with a separate bounded strategy.

## Test coverage

A focused test should verify that:

- stdout and stderr never exceed their configured byte limits;
- truncation flags are accurate for each stream;
- multibyte UTF-8 chunks do not bypass the byte limit;
- the child is still drained and reaches its normal exit status;
- output below the limit remains unchanged.

I can prepare the implementation if this approach fits the project's intended design.

deonmenezes:
Detailed and high quality description of the change, per this repo's own fork (not an external OpenAI/Codex contribution — this closes an issue the repo owner filed against `.codex/mcp-servers/`).

## What gap this closes

Closes #149.

`.codex/mcp-servers/lib/run_tool.js`'s `runCommand` is the shared subprocess helper for every Detect-stage MCP server: semgrep, CodeQL, osv-scanner, trufflehog, bandit, trivy, ast-grep, and z3 all spawn their underlying binary through it. It accumulated every stdout/stderr chunk into two plain JS strings with no size limit. A scanner that misbehaves — a runaway verbose/debug mode, a huge target repo, an infinite log loop — grows those two strings without bound until the MCP server process runs out of memory, which takes down *every* in-flight scan, not just the one that triggered it. That's a much worse failure mode than the "report, don't crash" posture the rest of this helper already follows for a missing binary (`notFound`) or a hung process (`timeoutMs` + `SIGKILL`).

## Why it's safe / why now

- It's the plumbing underneath the whole Detect toolchain, so a bug here silently threatens every scanner's reliability, not just one vuln class — exactly the "bug in the scanning pipeline" this routine looks for.
- Small, mechanical, fully backward compatible: no existing caller passes `maxBufferBytes`, so every current call site keeps its old behavior (default cap is a generous 20MB per stream — scanner JSON output is normally KB-MB). Only pathological output is affected, and the new behavior for that case is graceful truncation instead of an OOM crash.
- Once a stream crosses the cap, further chunks are dropped but the pipe keeps being drained, so the child process never blocks on backpressure and still exits normally — verified in the added test (`runCommand still drains and exits normally when the cap is hit`).
- Result now carries a `truncated` flag callers can check later if they want to surface "output was capped" (existing callers already tolerate/handle a JSON-parse failure gracefully, e.g. `semgrep/server.js`'s `catch` block, so no caller changes were required this round).

## Testing

Added `.codex/mcp-servers/lib/run_tool.test.js` using Node's built-in `node:test` (zero extra dependency, consistent with these servers' "pure Node" style) covering: output under the cap is untouched, stdout is capped near the limit with `truncated: true`, stderr is capped independently of stdout, the child still drains and exits well before the timeout when the cap is hit, and default behavior for normal-sized output is unchanged.

```
$ node --test .codex/mcp-servers/lib/run_tool.test.js
# tests 5
# pass 5
# fail 0
```

Also ran `npx prettier --check` on both changed files (repo's formatting config, `.prettierrc.toml`) — clean.

---

https://claude.ai/code/session_016zmSbrzFNCUnQPhaRbFJqc

---
_Generated by [Claude Code](https://claude.ai/code/session_016zmSbrzFNCUnQPhaRbFJqc)_


deonmenezes:
CI status on this diff (`.codex/mcp-servers/lib/run_tool.js` + new test only, no Rust/Cargo/npm-release files touched):

- ✅ `rust-ci / CI results (required)` — passes.
- ❌ `cargo-deny / cargo-deny` — fails on `advisories FAILED`. This is **not caused by this diff**: it fails identically on #180 (an unrelated docs-only PR), and it's still failing even on #177 (`fix/crossbeam-epoch-rustsec-2026-0204`), which specifically targeted a RUSTSEC advisory bump — so the advisory gate isn't fully fixed by any currently open PR I can port. This is a pre-existing, repo-wide Rust dependency-advisory issue outside this PR's scope (a pure JS change).
- ❌ `repo-checks / build-test` — fails trying to `gh run download` native npm-release artifacts from a pinned upstream `openai/codex` Actions run (`26201494185`): `no valid artifacts found to download`. Also reproduces identically on #180. This is a stale/expired pinned artifact reference in `scripts/stage_npm_packages.py`, unrelated to this diff and not something a re-run will fix (it's not a flake — the referenced upstream artifacts are gone, not transient).
- ⚠️ `Vercel` deployment check — unrelated to this repo's own code; it's reporting on a different, unrelated Vercel project (`opentradex-dashboard`) that appears to be wired to this GitHub account/org rather than to mantishack's own build.

Given none of the three failing checks touch code this PR changes, and no open PR currently produces a green `cargo-deny` or `repo-checks`, I'm not widening this PR to chase either — that's separate pre-existing repo maintenance. Flagging both to the repo owner instead. `rust-ci`, the check that actually reflects this diff, is green.

---

PR #184, opened 2026-09-19
Record blocked authorized-scan attempt against bounty target

deonmenezes:
## What this is

Not a detection-logic change — this is the scan-findings artifact for this run's discovery-scan leg against the authorized target `https://vulnerable-bounty-site.vercel.app`.

## What happened

Every attempt to reach the target failed at the TLS-CONNECT stage (before any HTTP request left the session): the session's outbound-HTTPS egress proxy returned `403 Forbidden` on `CONNECT vulnerable-bounty-site.vercel.app:443`, and its own status log recorded it as a `connect_rejected` / policy denial.

Per the proxy's documented guidance, a 403/407 on CONNECT means the destination host isn't on this session's egress allowlist, and the correct response is to report the blocked host rather than retry or route around it — so that's what this run did. No requests reached the target; there is no scan data (headers, body, injection probes, etc.) to report yet.

See `reports/security-scans/2026-09-19-vulnerable-bounty-site.md` for the full detail, including the exact curl/proxy-status evidence and a recommendation to add the target host (and the other 9 targets mentioned as forthcoming) to the environment's egress allowlist so future scheduled runs can actually reach it.

## Testing

N/A — documentation-only change (a new file under `reports/security-scans/`).

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01FvpacKYRcYWzi4WEXtKjP2

---
_Generated by [Claude Code](https://claude.ai/code/session_01FvpacKYRcYWzi4WEXtKjP2)_

PR #183, opened 2026-09-19
Add SSRF sink coverage to program-analysis source/sink heuristic

deonmenezes:
## What gap this closes

`mantis_program_analysis`'s `source_sink_scan` (the fast, dependency-free heuristic that tags attacker-controlled sources and dangerous sinks per `.codex/mcp-servers/program-analysis/source_sink_rules.js`) had rules for CWE-95 (eval/code injection), CWE-78 (command injection), CWE-79 (XSS), CWE-502 (insecure deserialization), and CWE-89 (SQL injection, Java only) — but **zero coverage for SSRF (CWE-918)**, even though the tool already tags the exact request-derived sources (`req.query`, `request.args`/`request.GET`/`POST`, `r.URL.Query()`, `request.getParameter`) that commonly feed an attacker-controlled URL into an outbound HTTP call.

This is a real coverage gap: a candidate like `axios.get(req.query.url)` would have its source (`req.query`) tagged but not its sink, so the pipeline had no signal to start tracing an SSRF path at all.

## What changed

Added SSRF sink rules (`cwe: "CWE-918"`) for the outbound-request call sites in each language `source_sink_scan` already supports:

- **JS/TS**: `axios.(get|post|put|delete|patch|request)`, `fetch(`, `http(s).(get|request)`
- **Python**: `requests.(get|post|put|delete|head|patch|request)`, `urllib.request.urlopen(`
- **Go**: `http.(Get|Post|PostForm|Head)(`
- **Java**: `new URL(...).openConnection()`

Same style as the existing sink rules: a plain call-site regex proxy (no argument/taint analysis), consistent with this file's documented role as a high-recall candidate surfacer for the Validate-stage agent to actually trace — not a dataflow engine.

## Testing

- `node --check` on the modified file.
- `npx prettier --check` (repo's `.prettierrc.toml`) — clean.
- Loaded `RULES` directly and unit-verified each new pattern matches its target call shape (`axios.get(userUrl)`, `fetch(target)`, `http.get(url, cb)`, `https.request(opts)`, `requests.get(url)`, `urllib.request.urlopen(url)`, `http.Get(url)`, `new URL(userInput).openConnection()`).
- Ran the heuristic end-to-end against a small fixture (`req.query` → `axios.get`, `request.args` → `requests.get`) and confirmed it now reports both the pre-existing source and the new SSRF sink in the same file/scan, where before only the source was surfaced.

[...]

PR #185, opened 2026-09-19
fix: widen mantis_http_audit inline-secret redaction coverage

deonmenezes:
## Summary

- Widens the `mantis_http_audit` MCP server's inline-secret redaction (`.codex/mcp-servers/http-audit/server.js`) beyond AWS/GitHub/Slack-token/JWT/private-key shapes to also catch: Google API keys (`AIza...`), OpenAI keys (`sk-...`), Stripe live keys (`sk_live_`/`rk_live_`), SendGrid keys (`SG....`), npm tokens (`npm_...`), Slack incoming-webhook URLs, and bare `Bearer <token>` values embedded in a captured body (as opposed to the `Authorization` header, which was already stripped).
- Adds the first automated test for this server (`server.test.js`, run with `node --test`): spawns it over stdio exactly as the harness does, calls `http_audit`, and asserts both the newly-added redactions and no regression on the pre-existing ones.
- Updates the `http-evidence` skill doc to list the widened coverage so agents know what's safe to rely on.

## Why

`mantis_http_audit` is the tool agents are told to route all captured DAST evidence through specifically so raw secrets never land in a finding, artifact, report, or prompt (PRD section 9/11 invariant, see `.codex/skills/http-evidence/SKILL.md`). The pattern list it used only covered a handful of shapes. Any of the very common key formats above — e.g. a leaked Google/Stripe/SendGrid/OpenAI key or webhook URL turning up in a response body while probing a target — would have passed through completely unredacted into the evidence pack attached to a finding, which is exactly the failure mode this tool exists to prevent. This closes that gap without touching the server's public API or behavior for anything it already redacted correctly.

## Test plan

- [x] `node --test .codex/mcp-servers/http-audit/server.test.js` — 3/3 passing (new-shape redaction, no-regression on prior shapes, non-secret bodies pass through untouched)
- [x] `npx prettier --check` on all touched files

https://claude.ai/code/session_01E5rNpsZhHac6cniQE2YGFt

---
_Generated by [Claude Code](https://claude.ai/code/session_01E5rNpsZhHac6cniQE2YGFt)_


deonmenezes:
`repo-checks / build-test` is failing, but it's not this PR's: the job fails at the "Download native artifacts" step in `scripts/stage_npm_packages.py`, which fetches from a **hardcoded** `WORKFLOW_URL = https://github.com/openai/codex/actions/runs/26201494185` pinned in `.github/workflows/repo-checks.yml` (line 58). That upstream run's artifacts have expired/been garbage-collected (`no valid artifacts found to download` for `x86_64-unknown-linux-musl`), so the step now fails unconditionally regardless of what any PR changes.

This is base-branch infra, not a flake — it's deterministic and reproduces identically across every recent open PR back to at least #176 (checked #176, #178, #180, #181, #183, #184: all show the identical `repo-checks / build-test` failure at the same step for the same reason), so I'm not spending the one re-run on it.

No fix for it exists yet in any open PR (I checked), so per the CI-red playbook I'm not widening this PR to carry an unrelated fix. Proposed fix for a follow-up: point `WORKFLOW_URL` at a current successful `openai/codex` rust-release run, or better, have `stage_npm_packages.py` resolve the latest successful run via the GitHub Actions API instead of a hardcoded run ID so this doesn't silently rot again.

Separately, worth flagging to @deonmenezes directly: this repo currently has ~85 open PRs, almost all drafts opened by this recurring maintenance routine since 2026-07-07, and none appear to have been merged since #106. Several close the same detection gap more than once (e.g. SQL-injection sink coverage: #101, #104, #114, #127, #139, #142, #152, #161, #167, #169) because the earlier fix never landed on `main`.

---

deonmenezes:
A second, distinct check is also red: `cargo-deny / cargo-deny`, failing on the `advisories` lane with `RUSTSEC-2026-0257` (Unix `BROWSER` handling allows browser argument injection) in `webbrowser` v1.0.6, pinned at `codex-rs/Cargo.lock:1123` via `codex-rs/Cargo.toml:455` (`webbrowser = "1.0"`, consumed by `login`, `rmcp-client`, `tui`). Also not this PR's: my diff is pure JS under `.codex/`, doesn't touch `Cargo.lock`/`Cargo.toml`, and this pin is already on `main`.

No open PR fixes this yet. Proposed patch, following the same shape as #177's `crossbeam-epoch` advisory bump: widen `codex-rs/Cargo.toml:455` to `webbrowser = "1.2"` (fix is in 1.2.2+) and update `Cargo.lock`.

---

PR #186, opened 2026-09-19
fix: exclude comments from source_sink_scan matches to cut false positives

deonmenezes:
## Summary

`source_sink_scan` (`.codex/mcp-servers/program-analysis/server.js`) is the heuristic regex matcher every detection-breadth PR builds on top of (sink/source pattern tables for JS/Python/Go/Java). It ran its regexes over **raw file bytes with no comment awareness**, so a pattern like `eval(`, `os.system(`, or `exec(` mentioned in a comment, changelog note, or docstring (e.g. `// don't use eval() here, it's dangerous`) was reported as a live sink/source candidate exactly like real code. Every prior detection PR has added more sink patterns on top of this same matcher without touching this false-positive path.

## Change

Added `blankComments(content, lang)`: a small per-language (js/go/java/py) lexer-lite pass that blanks out line/block comment regions (replacing with spaces, preserving offsets/newlines) before the sink/source regexes run. String and template-literal contents are deliberately left untouched — only tracked, so a `//` or `#` inside a string (e.g. a `https://` URL) isn't mistaken for a comment start, which would otherwise truncate the rest of the line from matching. This means no new false negatives are introduced: any candidate that would have matched before still matches, minus the comment-only mentions.

Wired into `sourceSinkScan` in place of scanning raw `content`; snippet/line-number extraction still comes from the original file text so reported snippets are unchanged.

## Testing

No test harness exists yet for the `.codex/mcp-servers/*` Node servers (none of the prior sink-coverage PRs added one either), so this was verified manually:
- `node --check` on the modified file.
- `npx prettier --check` passes (repo formatting).
- Manual fixture run (JS): a same-line `//` comment, a multi-line `/* */` block comment, and a docstring-style comment mentioning `eval(`/`exec(` are now correctly excluded from `sinks`; a `https://example.com//eval(fake)` string literal is *not* truncated by the `//`; real code (`eval(req.query.x)`, `child_process.exec(req.body.cmd)`, `document.write(location.hash)`) is still flagged.
- Manual fixture run (Python): a `# ...` line comment mentioning `os.system(` is excluded; a triple-quoted docstring mentioning `os.system(` is still flagged (correct — Python docstrings are string literals, not comments, so this is intentionally unchanged); real `os.system(cmd)` usage is still flagged.

## Scope note

This is part of the repo's recurring automated detection-improvement routine. Deliberately picked a false-positive fix over another sink-category addition — the repo currently has a large number of open, unmerged PRs proposing near-duplicate sink-coverage additions (SQLi, SSRF, path traversal, XXE, etc.) on top of this same unfixed matcher; this closes a shared precision gap instead of adding to that pile.

---

https://claude.ai/code/session_01VKHgVJADxQK8EdXQ9T3xBi

---
_Generated by [Claude Code](https://claude.ai/code/session_01VKHgVJADxQK8EdXQ9T3xBi)_


deonmenezes:
CI on this PR's head is red on two checks, but neither is caused by this diff (which only touches `.codex/mcp-servers/program-analysis/server.js`):

**1. `repo-checks / build-test`** — [job](https://github.com/deonmenezes/mantishack/actions/runs/35455248623/job/105929257451) fails downloading pinned native release artifacts. `.github/workflows/repo-checks.yml` hardcodes `CODEX_VERSION=0.133.0-alpha.4` / `WORKFLOW_URL=https://github.com/openai/codex/actions/runs/26201494185`; `gh run download` now returns "no valid artifacts found to download" for that run — almost certainly GitHub's Actions artifact retention window expiring for it, not anything in this diff. Confirmed this reproduces identically on PR #185 (unrelated diff, opened ~3h before this one, same error text). No existing PR or commit in this repo fixes it — the pin was introduced at the initial `d155094` rebuild commit and hasn't been touched since.

**2. `cargo-deny / cargo-deny`** — [job](https://github.com/deonmenezes/mantishack/actions/runs/35455248623/job/105929257401) fails on the `advisories` check against the locked Rust dependency graph (`Cargo.lock`), which this PR does not touch. Also reproduces identically on PR #185.

Both are pre-existing/base-branch issues, not regressions from this change, so I'm not widening this PR's scope to chase either one. Proposed fixes for whoever picks these up:
- For (1): repoint `WORKFLOW_URL`/`CODEX_VERSION` in `repo-checks.yml` at a currently-live `openai/codex` release run with retained artifacts, or make the artifact-download failure non-fatal for this PR-time check specifically.
- For (2): identify and bump/patch the flagged crate in `Cargo.lock`, or add a scoped `cargo-deny` ignore with a tracking note if no fix is available upstream yet.

---

https://claude.ai/code/session_01VKHgVJADxQK8EdXQ9T3xBi

---

deonmenezes:
Follow-up with the precise root cause for the `cargo-deny / cargo-deny` failure (item 2 above), found by reading the full job log rather than just the tail:

```
ID: RUSTSEC-2026-0258
Advisory: https://rustsec.org/advisories/RUSTSEC-2026-0258
Crate: h2 v0.4.13 (registry+https://github.com/rust-lang/crates.io-index)
Severity: Low
The h2 crate, used internally by hyper, had a flaw that would accept and queue empty
DATA frames without limit. If streams were not actively drained, this could lead to
unbounded memory usage, or a panic if the length overflows.
Patched in v0.4.16.
Solution: Upgrade to >=0.4.16 (try `cargo update -p h2`)
```

`h2` is a transitive dependency (pulled in via `aws-smithy-http-client` → `aws-smithy-runtime` → `aws-config` → `codex-aws-auth` → ... → most of the workspace), so this should be a `cargo update -p h2` away — no `Cargo.toml` edits needed. Per this repo's `AGENTS.md`, any `Cargo.lock` change also requires `just bazel-lock-update` to refresh `MODULE.bazel.lock` (CI checks for lockfile drift), and neither `just` nor `bazel` are installed in the environment this automated PR runs in, so I'm not attempting that lockfile refresh myself here — installing a full Bazel toolchain to do one workspace-wide dependency resolution is disproportionate to run speculatively in this loop, and a half-completed lockfile update would just trade one CI failure for another. Leaving this as a precise, low-effort pointer for whoever has `just`/`bazel` available: `cargo update -p h2` then `just bazel-lock-update`, commit both lockfiles.

---

https://claude.ai/code/session_01VKHgVJADxQK8EdXQ9T3xBi

---

PR #187, opened 2026-09-19
fix: close secret-shape gap in mantis-findings DLP refusal guard

deonmenezes:
## What gap this closes

`.codex/mcp-servers/findings/server.js` is the tool-owned findings spine (PRD section 9/11): every `finding_create`/`finding_update` call is supposed to be refused if the payload "looks like it contains a raw secret/token/key" (see `.codex/skills/findings-spine/SKILL.md`, which documents the guard as covering "a raw secret, token, key, or JWT" in general).

In practice, `scanForSecrets()`'s pattern list only covered five specific shapes: AWS access key IDs, GitHub tokens, Slack tokens, PEM private keys, and JWTs. Its sibling tool, `.codex/mcp-servers/http-audit/server.js`, redacts a *broader* set of shapes out of evidence packs — most importantly generically-named secret params (`token=`, `api_key=`, `password=`, `session_id=`, ...) and `user:pass@` credentials embedded in a URL.

That meant a raw secret in one of those generically-named/URL-embedded shapes could still land, unredacted, in the append-only `.codex/findings/events.jsonl` log via a hand-written `claim`/`attack_vector`/`evidence` field that never passed through `http_audit` first — exactly the kind of leak the findings spine's refusal gate exists to catch.

## What changed

- Added the missing two pattern shapes to `SECRET_PATTERNS` in `server.js`, bringing it to parity with `http-audit`'s redaction list: generically-named secret/token/password params, and `user:pass@` URL credentials.
- Used non-global regexes deliberately: `server.js` calls `.test()` on these patterns, which is stateful (advances `lastIndex`) on a `g`-flagged `RegExp` and would make the guard flap between catching and missing the *same* input across successive calls.
- Deliberately did **not** port `http-audit`'s bare `auth` keyword into this list. Over there it just redacts a substring out of an evidence preview (cheap false positive). Here it would refuse the *entire* write, and `auth` shows up constantly in legitimate auth/access-control finding claims (e.g. "admin endpoint reachable without auth check") — porting it verbatim would have made the guard actively hostile to the exact class of findings this tool exists to record.
- Added `test_secret_guard.js`, a zero-dependency smoke test (no test framework in this repo for the `.codex/mcp-servers/*` servers) that spawns the real server over its MCP stdio transport and exercises: a clean finding (accepted), a generic `password=` param (now refused), `user:pass@` URL creds (now refused), a legitimate auth-related claim with no secret value (still accepted — guards against over-blocking), and a JWT (still refused — regression check). It runs the server from an isolated temp-directory copy so it can't write test fixtures into the real, gitignored `.codex/findings/events.jsonl` production log.

## Testing

```
$ node .codex/mcp-servers/findings/test_secret_guard.js
ok: clean finding_create is accepted
ok: generic password= param is refused
ok: user:pass@ URL credentials are refused
ok: legitimate auth-related claim (no secret value) is still accepted
ok: JWT-shaped string is still refused
```

- `node --check` on both changed/new files.
- `prettier --check` on both changed/new files (repo's configured formatter) — clean.

## Scan report (task 2 of this run)

This routine also runs a discovery-only scan against `https://vulnerable-bounty-site.vercel.app` each cycle. This run, that step could not execute: the session's network egress proxy explicitly blocks that domain (`EGRESS_BLOCKED` from the web-fetch path; the raw HTTP client got a `403` on the CONNECT tunnel). No scan traffic was sent and no findings were fabricated. This looks like an environment/network-policy setting rather than anything about the target itself — flagging so the environment's egress allowlist can be adjusted if live scanning against that target is meant to run from this session in future cycles.

---

https://claude.ai/code/session_0173aPr5kVEdWCYdEfo7M6o8

---
_Generated by [Claude Code](https://claude.ai/code/session_0173aPr5kVEdWCYdEfo7M6o8)_


deonmenezes:
Checked CI on this PR's head (`99fae1b`). Three checks are red, but none are caused by this diff (which only touches `.codex/mcp-servers/findings/server.js` and adds `.codex/mcp-servers/findings/test_secret_guard.js`):

1. **`repo-checks / build-test`** — fails in `scripts/stage_npm_packages.py` trying to download native artifacts from a hardcoded, now-expired `openai/codex` Actions run (`WORKFLOW_URL=https://github.com/openai/codex/actions/runs/26201494185`, `CODEX_VERSION=0.133.0-alpha.4`): `no valid artifacts found to download`. Nothing in this diff touches packaging/native artifacts.
2. **`cargo-deny`** — fails on `advisories`: `RUSTSEC-2026-0257` (Unix `BROWSER` argument-injection vulnerability) in `webbrowser 1.0.6`, pinned in `codex-rs/Cargo.lock`. Fix upstream is bumping `webbrowser` to `>=1.2.2`. That's a Rust dependency bump across the `codex-rs` workspace — out of scope for this PR's two-file JS fix, but worth a follow-up (this repo scans for exactly this class of issue, so it'd be good to not carry it in its own lockfile).
3. **`Vercel`** (`opentradex-dashboard`) — deployment failure for a project that has no presence in this repository at all (no `vercel.json`, no matching source, `grep -ri opentradex` finds nothing). Looks like an account-level Vercel GitHub App integration pointed at the wrong project; not something a code change here can fix.

Confirmed (1) and (2) are pre-existing on `main`, not introduced by this PR: `blocking-ci` on the last 5 pushes to `main`, including the current tip (`c3a2e68`), all show `conclusion: failure` or `cancelled` — same `repo-checks`/`cargo-deny` shape. No re-run needed; this isn't a flake.

I haven't pushed anything for these since none are in this PR's diff and (2)/(1) both need changes well outside its scope. Flagging here so a human can decide whether to open follow-up issues for the stale artifact reference and the `webbrowser` bump.

---

PR #190, opened 2026-09-20
fix: redact JSON-bodied secrets in http-audit evidence packs

deonmenezes:
## Summary

- `mantis_http_audit`'s generic secret-param redaction (`SECRET_VALUE_PATTERNS` in `.codex/mcp-servers/http-audit/server.js`) only matched query/form `key=value` pairs (e.g. `token=abc123`). A secret carried as a JSON body field — `{"access_token": "abc123"}` — is the dominant shape for modern REST API traffic, and it passed through into the evidence pack completely unredacted.
- Added a JSON-shaped (`"key": "value"`) variant of the same generically-named-secret rule (token/api_key/access_token/refresh_token/secret/password/passwd/auth/session_id/sig/signature), so JSON response/request bodies get the same coverage query strings already had.
- Also closed a few other coverage gaps that had zero patterns before: Google API keys (`AIza...`), Stripe keys (`sk_live_`/`pk_live_`/`rk_live_`/`sk_test_`/etc.), SendGrid keys (`SG.xxx.yyy`), and npm tokens (`npm_...`).

This closes a real gap in the Detect/Validate pipeline's own DLP layer: `http_audit` is the tool agents use to attach DAST evidence for a finding (see `.codex/skills/http-evidence/SKILL.md`), and its whole job is to guarantee secrets never leak into a finding/evidence/report. Since most APIs return JSON, the previous query-string-only pattern was a significant false-negative for exactly the traffic shape this tool sees most.

## Why this is safe / scoped

- Pure addition to an existing regex-based redaction list; no behavior change to anything that was already being redacted (verified with a new test asserting non-secret JSON fields like `"user":"alice"` are left untouched).
- Zero new dependencies; still a pure, offline Node function (no network, no external binary).

## Test plan

- [x] Added `.codex/mcp-servers/http-audit/server.test.js`, a black-box test that drives the server over its real stdio JSON-RPC transport (same path a Mantis agent uses).
- [x] `node --test .codex/mcp-servers/http-audit/server.test.js` — 4/4 passing:
  - redacts a secret carried as a JSON response body field
  - redacts recognizable secret shapes (Google API key, Stripe key)
  - still redacts the pre-existing query/form `key=value` shape (regression guard)
  - leaves non-secret JSON fields untouched
- [x] `npx prettier --check` clean on both changed files.

---

https://claude.ai/code/session_01P9yoppk1N4qZCdw9EPHSdU

---
_Generated by [Claude Code](https://claude.ai/code/session_01P9yoppk1N4qZCdw9EPHSdU)_


deonmenezes:
`repo-checks / build-test` is failing on this PR's head (5211ffe), but the failure is unrelated to this diff.

The job fails in the "Stage npm package" step, inside `scripts/stage_npm_packages.py`, which downloads pre-built native binaries from a **hardcoded upstream workflow run**: `.github/workflows/repo-checks.yml` pins `CODEX_VERSION=0.133.0-alpha.4` / `WORKFLOW_URL="https://github.com/openai/codex/actions/runs/26201494185"`. That `gh run download` call now fails with `no valid artifacts found to download` — almost certainly because that specific upstream Actions run's artifacts have since expired (this pin hasn't been touched since the July 6 rebuild commit, ~2.5 months ago, well past typical Actions artifact retention windows).

This PR only touches `.codex/mcp-servers/http-audit/server.js` and its new test — nothing in `scripts/`, `.github/workflows/`, or the npm packaging path. This same failure would reproduce identically on any current PR against `main`, including a no-op one.

No fix exists in this repo yet. The real fix is to re-point `CODEX_VERSION`/`WORKFLOW_URL` in `.github/workflows/repo-checks.yml` at a current, still-live `openai/codex` rust-release run — I don't have access to the `openai/codex` repo from this session to identify a currently-valid one, so I'm not bundling a guess into this PR. Re-running the failed job once to confirm this isn't transient.

---

deonmenezes:
`cargo-deny / cargo-deny` is also failing on this PR's head, and it's likewise unrelated to this diff.

`cargo-deny`'s `advisories` check fails on **RUSTSEC-2026-0257** (https://rustsec.org/advisories/RUSTSEC-2026-0257): `webbrowser` v1.0.6 has a Unix `BROWSER`-env argument-injection vulnerability, pulled in transitively via `codex-login`. This PR doesn't touch `Cargo.toml`, `Cargo.lock`, or any Rust code — only `.codex/mcp-servers/http-audit/server.js` and its test — so `codex-rs/Cargo.lock` here is byte-identical to `main`'s, and this same advisory would fail identically on `main` itself or any other current PR.

No fix exists in this repo yet. Per the advisory, the fix is a one-line dependency bump (`cargo update -p webbrowser` to `>=1.2.2`), but that's a real (if small) change to `codex-rs/Cargo.lock` across the whole Rust workspace, unrelated to and out of scope for this PR's JS-only fix — flagging it here rather than bundling it in.

---

